PR #36, opened 2014-07-10
Refactored processingHeaderBlock in HeaderParser.Java

philgooch:
It would be useful to have the option to serialize the PDF header in a format other than TEI XML, e.g. JSON. I have extracted out the header object creation from processingHeaderBlock() into a new getBiblioItem() method, so that a Grobid client can pass the header to a different formatter if required.


lfoppiano:
I close this PR as it's quite old

PR #52, opened 2015-04-09
Indentation feature for ReferenceSegmenter

rodneykinney:
Thanks for the guide through the code!

### DIFF
--- a/grobid-core/src/main/java/org/grobid/core/engines/ReferenceSegmenterParser.java
+++ b/grobid-core/src/main/java/org/grobid/core/engines/ReferenceSegmenterParser.java
@@ -676,18 +676,21 @@ System.out.println("");
 
                 LayoutToken token = null;
                 if (tokens != null) {
-                    while (tokenIndex < tokens.size()) {
-                        token = tokens.get(tokenIndex);
-                        if (text.equals(token.getText()))
+                    int i = tokenIndex;
+                    while (i < tokens.size()) {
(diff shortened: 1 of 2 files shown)


grobid-core/src/main/java/org/grobid/core/engines/ReferenceSegmenterParser.java
```diff
@@ -636,68 +644,80 @@ else if (n == block.getEndToken()) {
...
-						if (text.equals(token.getText()))
-							break;
-						tokenIndex++;
-					}
-				}
-				
+
+                LayoutToken token = null;
+                if (tokens != null) {
+                    int i = tokenIndex;
```
rodneykinney: Improve robustness of matching to LayoutToken

grobid-core/src/main/java/org/grobid/core/engines/ReferenceSegmenterParser.java
```diff
@@ -636,68 +644,80 @@ else if (n == block.getEndToken()) {
...
+                            break;
+                        }
+                        i++;
+                    }
+                }
+
                 if (previousNewline) {
                     newline = true;
                     previousNewline = false;
+					if (token != null && previousFeatures != null) {
```
rodneykinney: The indentation logic.  Compare start of line to previous line, change indentation if the difference is more than a character width.

PR #54, opened 2015-06-12
istex mods to grobid

rloth:
Hello Patrice !

I've been using these two changes internally and now I submit them for your opinion:
  - on the training, commit e33b9a7 provides needed rawtxt files for my gold-generating process "ragreage"
  - for the tagging itself, commit 674178b proposes a new route to query the service with a param containing a remote pdf url (needed to query our API without additional copying of the file)

### DIFF
--- a/grobid-core/src/main/java/org/grobid/core/engines/Engine.java
+++ b/grobid-core/src/main/java/org/grobid/core/engines/Engine.java
@@ -283,6 +283,15 @@ public class Engine implements Closeable {
                     writer = new OutputStreamWriter(new FileOutputStream(new File(outPath), false), "UTF-8");
                     writer.write(raw + "\n");
                     writer.close();
+
+                    // ----------------------8<-----------------------------
+                    // rloth: also write the raw text as it is before refseg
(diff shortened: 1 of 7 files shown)


kermitt2:
Thank you Romain !

About querying the service with a parameter containing a remote pdf url:

- this kind of query with PDF upload was actually in the old version of GROBID (circa 2011), but I removed it for possible security issues 

- the proxy data are already defined in GROBID in the property file. The REST API permits to modify these property values (see the administration part). I don't think it is something to be specified directly in the REST query because it is really a set up data for the server.

kermitt2:
The additional generation of files with raw text when creating training data have been merged in Grobid with commit 6a658bff5bbf77b74f1fef4626bbe316a881cdcb

rloth:
Hello
Thanks for adding the generation of rawtexts and the xml entity thingy in the master !

As for the rest, I'll keep the remote-pdf route in our own downstream version along with our trained crf models... (but i'll correct the remote-pdf thing to take config's grobid.proxy_host into account).

I'm closing the pull request !

PR #60, opened 2015-07-31
INSPIRE-related extensions

jcboyd:
- Added training data
- Modelling collaborations as a class of the header model


kermitt2:
Hi Joseph!

Thanks a lot for your contribution. May I ask you a couple of questions?

* what is exactly a _collaboration_? I have the impression that it is a particular type of affiliation, e.g. 
```xml
       <affiliation>The LUNA Collaboration<lb/></affiliation>
```
So it would not be part of the header model, but part of the affiliation model as an additional type of "organisation" (in addition to laboratory, department and institution, cf. the encoding guidelines for affiliations). Adding it to the affiliation model is much easier than in the header model. 

* I just looked rapidly at the new training files, in the new header training data I saw something like that:
```xml
       <orgName>DELPHI Collaboration<lb/></orgName>
```
```<orgName>``` is not a valid tag for the training data of header, it should be 
```xml
<byline><affiliation>... </affiliation></byline> 
```
Or is it how you wanted to encode the _collaboration_ field? 

* I've seen some nice new INSPIRE resources files :) 
I will look at them to see if I can add them easily when the Lexicon object is created, then these resources could be used in the CRF models immedialty (after retraining). 

* I have the impression that all the new training file for reference segmentation follow exactly the same layout style? Is it correct? 
In this case, we have to be careful not to create biased training data by representing too much one layout style. I will try to do some tests.

Thanks again for your great contribution!

jcboyd:
Hi Patrice!

In the HEP world a collaboration is viewed as something between an affiliation and an author. For example, the current world record holder for authorship (http://arxiv.org/abs/1503.07589), combines two CERN collaborations, ATLAS and CMS, which represent over 5,000 physicists (all of whom are listed in the rear of the paper). This is why I was modelling them as a separate field, but I can see how this could be more easily modelled in the affiliation model.

I was experimenting with the INSPIRE resources as dictionary-based features and they seem to give a performance increase for HEP papers. I also found that a dictionary of stop words gave an improvement.

Regarding the reference-segmenter data, I fully agree this set is too homogeneous to train a general model. It was part of a separate experiment and I should have excluded it from the pull request.

Thank you for your feedback! We may have more questions when GROBID goes "live" in INSPIRE :)

kaplun:
cc: @jalavik, @kaplun, @jacquerie

lfoppiano:
Hi, I've merged your PR in a separate branch https://github.com/kermitt2/grobid/tree/jcboyd-pykelet-PR-segmentation. Me and @kermitt2 we will validate then the training data (the format has changed so we might need to adapt), we will integrate the change. 

After that I would suggest you to consider updating your grobid version to the latest (I see the https://github.com/jcboyd/grobid is pretty old). 

I apologise for such a long delay, thanks again for your contribution. ;-)

kaplun:
@lfoppiano this is great news. At INSPIRE we are more and more using Grobid within our workflows and this will greatly help :+1:

lfoppiano:
Hi guys, 
     sorry for the long waiting. We are moving to integrate this PR in Grobid. 
Let start from the reference-segmenter and segmentation model. For both this model we would like to regenerate the feature files, and to do that we would need the original pdf. 

In the files that you provided there are different IDs. For the arXiv Ids (e.g. 1309.7222.training.segmentation.tei.xml) we can download automatically the data but for the other (e.g. 1344707.training.segmentation.tei.xml) we can get the page from INSPIRE, but, in this case there is no direct link to the pdf (opposite case for the ArXiv pointed files). 

Would you be able to provide us the original PDFs or a mapping to the open archive Ids?  

Thanks

kaplun:
Hi @lfoppiano,
Yep, I will prepare for you a zip file with most of the original sources that have been used for this PR. I hope to pick up the correct version (sometimes there are more than one).

lfoppiano:
This PR is quite obsolete, can be closed

PR #63, opened 2015-08-10
improve handling of HTML entities

fmux:
Dear Grobid team,

I had some trouble with Grobid not properly escaping XML special characters in the `xml:id` attribute when generating `biblStruct`s.

Also I noticed a typo in the part of the `TextUtilities.HTMLEncode` method that tries to prevent double encoding of HTML entities (it was checking for `&quote;` instead of `&quot;`.

Let me know what you think!

All the best,
Fabian


kermitt2:
Hi Fabian,

Many thanks for your fixes! This is all very absolutely correct and correct my errors. 

All the best!

Patrice

PR #137, opened 2016-10-03
Rework the way stats are calculated

boumenot:
This may be too big of a change...

I was trying to understand how Grobid evaluates a model, and the best way I found to do that was to refactor the code.  It produces the same results in my *limited* testing.

### DIFF
--- a/grobid-core/src/test/java/org/grobid/core/engines/EngineTest.java
+++ b/grobid-core/src/test/java/org/grobid/core/engines/EngineTest.java
@@ -1,17 +1,13 @@
 package org.grobid.core.engines;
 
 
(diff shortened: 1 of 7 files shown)


boumenot:
No need to apologize.  I too have been busy, and I haven't even addressed the issue caught by CI.

kermitt2:
Apparently two classes missing in the PR - LabelStat.java and Stats.java

boumenot:
Gah!  I'll commit them when I get home.

kermitt2:
Tested on the PMC sample set (~2000 documents) and same results, let's merge, thanks !

PR #144, opened 2016-10-20
Add an alternative name for China to country codes

qtux:
It would be useful to add 'Peoples R China' as an additional English alternative name to the country code lexicon. The extracted affiliation strings from Web of Science seem to contain this name quite often.


kermitt2:
Thanks !

PR #145, opened 2016-10-21
add alternative country names to country codes

qtux:
It would be useful to add the alternative names for
- Belarus
- Ivory Coast
- United Arab Emirates
- Papua New Guinea
- Reunion
- Northern Ireland

to the country code lexicon. The extracted affiliation strings from Web of Science seem to contain this names in some cases.


kermitt2:
Thanks!

Issue #155, opened 2017-01-04
How to get GROBID version with RESTfull service


Katharsis:
Hi guys,

Is there any way for getting GROBID version using REST API?

GET: http://<domain.com>/version/

On output I'd like to see text/plain or JSON with string "GROBID 0.4.1".
It will be very helpful for me.

lfoppiano:
Dear @Katharsis thanks for your email, 
    indeed this feature is not present currently. We can add it. ;-) 

If you have already an idea on how to tokenize the version with maven, feel free to propose it ;-)

Cheers
Luca

Katharsis:
@lfoppiano 

I'm not a Java developer but take a look at here:

https://web.archive.org/web/20120121133235/http://www.christophkiehl.com/easy-way-to-display-your-apps-version-using-maven-and-manifest

Is that what you asking for?

lfoppiano:
@Katharsis Thanks. I'll have a look. ;-)

lfoppiano:
Thanks to @czuraj for the PR. 
@Katharsis if you can test it would be great ;-)

PR #156, opened 2017-01-06
Added RESTfull endpoint for getting grobid version

czuraj:
This PR should resolve https://github.com/kermitt2/grobid/issues/155


lfoppiano:
After rewvieing this PR I've made several changes. 
I think getting the hardcoded or tokenised version from the property file grobid.properties is not a viable solution, because as it is designed at the moment, it's filtered/tokenised only when is included in a jar/war. I've managed to tokenize it correctly using the assembly plugin in the grobid-home project, but that's not satisfying.

I've put this information in the web.xml where is tokenised from maven when the project is built. This will allow us to test properly the feature by avoiding changing of values in the tests when the version changes.

lfoppiano:
FYY I've also changed the entry point from /getVersion to /version, as the GET is implicit in the method used to call the endpoint. ;-) 

Please feel free to test it and let me know ;-) it's in master now.

PR #157, opened 2017-01-06
Added Maven wrapper

krzysiekruczaj:
I use Gradle as my build tool, so I haven't got Maven installed on my computer. That's why it would be ideally to include Maven Wrapper from https://github.com/takari/maven-wrapper to make builds using one and stable version of Maven.


lfoppiano:
@czuraj I think there is a gradle build for grobid, but I'm afraid is not up to date. 
I personally would prefer to bring the gradle build up to date with the maven one rather than adding an additional layer on maven. 

In your case you would just use gradle ;-) What do you think of this?

lfoppiano:
Since we are moving to gradle, I think this PR is not needed anymore. Thanks anyway

PR #183, opened 2017-05-09
Fix for the May/March ambiguity.

aoboturov:
There are two things in this commit:

1. Patterns are simplified to change (J|j) to [Jj] which accordingly to Intellij should work faster :)
2. Months patterns are matched till the end of string, this would prevent the test "testMayAndMarchOverlap" to fail for *May* because it would match *ma* for Pattern *mar* first.


kermitt2:
Many thanks for the PR !

PR #204, opened 2017-07-18
Documention of article IDs for reference training

michamos:
I replaced 
> **`<biblScope type="pp">`** the full range of pages

by

> **`<biblScope type="pp">`** the full range of pages or the article ID


kermitt2:
many thanks !

PR #205, opened 2017-07-21
Dropwizard service

detonator413:
- Reworked how Grobid Home is found now
   - Search in classpath
   - Search via property (able to provide http/https path)
   - Locally around the working directory
-Dropwizard integration 
  - Adapted JS files
  - Dependency injection
  - Exception mapper so that it's not necessary to construct a response object for failed requests


lfoppiano:
As soon as we have done the new grobid release, I think we can merge the dropwizard on the master. However I would take a later version of the branch which I'm keeping up-to date with the master, in the meantime.

lfoppiano:
When I run tests using the `--info` I've get this: 
```
    SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
    SLF4J: Defaulting to no-operation (NOP) logger implementation
    SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.
```
All logs implementation are excluded in the gradle.build. AFAIK we need to provide a log4j/slf4j implementation in grobid-core (Artem set it up as runtime - it's needed for tests and batch processing), grobid-training and grobid-service as well.

kermitt2:
Actually it's me who add it as runtime for grobid-core, Artem removed the implementation completely at the time resulting in the same problem for batch processing. It was already specified separately for the test in maven, and for each pom of the other sub-project.

PR #207, opened 2017-08-01
Fix 'zh-tw' language detection.

aoboturov:
Exception occurs for some documents:

    org.grobid.core.utilities.LanguageUtilities: Cannot detect language because of: 
    org.grobid.core.exceptions.GrobidException: [GENERAL] Language id should consist of two or three letters, but was: zh-tw


kermitt2:
Many thanks !

PR #212, opened 2017-08-02
Remove slf4j implementation from the Grobid Core.

aoboturov:
SLF4j implementation should not be a part of the Grobid Core - the implementation is provided by the user of the library.


aoboturov:
@kermitt2 Could you check this one please?

kermitt2:
All good thanks ! I was just checking if it would had an impact on the other grobid modules/clients

PR #216, opened 2017-08-02
GrobidDefaultAnalyzer.tokenize should ignore TextUtilities.spaceDelim…

### LINKED ISSUE #214: tokenization produces corrupted features vectors for fulltext
C++ isspace() would treat \r\t\n as symbols which would be removed in Wapiti's `src/reader.c`:

		uint32_t cnt = 0;
		while (*line != '\0') {
			toks[cnt++] = line;
			while (*line != '\0' && !isspace(*line))
				line++;
			if (*line == '\0')
				break;
			*line++ = '\0';
			while (*line != '\0' && isspace(*line))
				line++;
		}

We have PDF2XML which actually produces such sequences: `´\rÓÑÔÙØØØ`. In return, it would be set up to the text value of org.grobid.core.layout.LayoutToken. And this text would be used to serialize the the org.grobid.core.features.FeaturesVectorFulltext.printVector()

aoboturov:
…iters fixes #214.

### DIFF
--- a/grobid-core/src/main/java/org/grobid/core/engines/ReferenceSegmenterParser.java
+++ b/grobid-core/src/main/java/org/grobid/core/engines/ReferenceSegmenterParser.java
@@ -1,5 +1,6 @@
 package org.grobid.core.engines;
 
+import com.google.common.base.Function;
 import com.google.common.collect.Sets;
 import org.grobid.core.GrobidModels;
 import org.grobid.core.document.Document;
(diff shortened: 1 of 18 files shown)


kermitt2:
Normally space should be preserved to make possible further processing/heuristics. 

```
text = text.replaceAll(TextUtilities.spaceDelimiters, "");
```

It's used in particular to recreate the original strings in the TEI output with actual space/no space.

aoboturov:
Could you check the space constant? The thing is that such values are
ignored by wapiti parser. So if you would like to keep them anyways, we
should clean them at the place which produces output for wapiti. What would
you say about such approach?

kermitt2:
Well that's how it works currently - separators are removed when creating the input for Wapiti, and then added back based on the result and the reference tokenization produced by the analyser. That's why it's important to preserve the separators in the tokenization. 

I'll check the separator constants at the place of creation of the Wapiti input - however the drawback it that this piece is repeated in each taggers class :/

aoboturov:
I think I got you idea - sanitization was applied to all features.

kermitt2:
I think this is going to work well thanks ! There's just one point I am not sure to see, how do we check if the prefix/suffix is empty after wapiti sanitization (example bellow), we might have an empty string, and thus an error in the number of features of the vector and a Wapiti failure :

```
 res.append(" " + TextUtilities.prefix(wapitiSanitizedString, 1));
```

aoboturov:
@kermitt2 normally

    if (wapitiSanitizedString == null) return null;
    if (wapitiSanitizedString.length() == 0) return null;

does the trick. If after sanitization the string is empty, it would not even be considered for wapiti input.

lfoppiano:
as a matter of completeness you can also use 

`if(StringUtils.isBlank(wapitiSanitizedString)) return null;`

:-)

aoboturov:
@lfoppiano done.

aoboturov:
The following document crashes ReferenceSegmenterParser:

[AS_380334165250048@1467690139278_content_1.pdf](https://github.com/kermitt2/grobid/files/1200637/AS_380334165250048.1467690139278_content_1.pdf)

Which was fixed by using the TaggingTokenSynchronizer.

aoboturov:
this code was integrated via the https://github.com/kermitt2/grobid/pull/218

PR #240, opened 2017-09-13
Recompile wapiti on macOS sierra to fix build issues

### LINKED ISSUE #142: Invalid Format Error in OSX
Problem running grobid on osx (actually macOS).

I used the installation process described in the doc :

```
wget https://github.com/kermitt2/grobid/archive/grobid-parent-0.4.1.zip
unzip grobid-grobid-parent-0.4.1.zip
mvn clean install
```

And then the tests don't pass, because _"invalid format error"_ when loading the wapiti models. If skip the test I get the same error when I try grobid on a PDF.

iorala:
I recompiled wapiti from https://github.com/kermitt2/Wapiti on macOS sierra. 
This fixes #142 for me. I don't need to set export LC_ALL="C" anymore.

It would be great if you could test it on your installations. 

I can't test if the library will work on older versions of macOS as I only work with sierra.


lfoppiano:
Hey @iorala, thanks for this PR and sorry it took so long to react to it. I'm going to find some time in the next weeks to test it and merge it. 

Regards
Luca

PR #246, opened 2017-10-06
Improved gradle setup. Removed unnecessary artifact creation in root …

Hillkorn:
…project and fixed jacoco report. Updated gradle wrapper to latest 4.2.1.


detonator413:
I merged the best parts (excluded local maven libs since it's too verbose, more places + didn't work for me for some reason)

Also lucene was shaded only in the onejar, but for dependent project we needed it to be shaded in the main jar. I added it too

Issue #264, opened 2017-11-24
Docker image for 0.5.0 version


mastak:
What about docker image for the new version?

There is `Dockerfile` in the repository,  but it requires already built files. 
Docker multi-stage build (https://docs.docker.com/engine/userguide/eng-image/multistage-build/) can be used here for.

lfoppiano:
Hi @mastak,  
    it's on my todo list :-) Many things has changed since the last release and it will require some re-thinking. If you have experience with Docker and you like to propose some solution, feel free. 

I'll try to get it done soon, however cannot guarantee when

Cheers
Luca

mastak:
Hi @lfoppiano,

I'm not very familiar with Java, but I can help with docker. 
I see that there are should be 2 files, but how I can build them from source?:
```
ADD ./grobid-home/target/grobid-home-${GROBID_VERSION}.zip /opt
COPY ./grobid-service/target/grobid-service-${GROBID_VERSION}.war /var/lib/jetty/webapps/ROOT.war```

lfoppiano:
@mastak you need to know a bit of building on java actually, we don't have a war build anymore, but I'm not sure it's perhaps good to add it back for the docker image. 
In the meantime if you use the docker image with tag 0.4.4 you will have roughly the same features, in version 0.5.x we have changed build method (gradle) and we added a framework. 

Cheers
Luca

lfoppiano:
@mastak with long delay, the Dockerfile is now updated (thanks for your contribution!) and the image is uploaded on the docker hub (https://hub.docker.com/r/lfoppiano/grobid/tags/)

PR #265, opened 2017-11-27
added multi step dockerfile for 0.5.0 version

mastak:
#264


lfoppiano:
I've check this out, we need to bring back the war build #272 as the image now it's 3Gb against 700mb it was before... ;-)

PR #306, opened 2018-04-19
[wip] build docker from local source

de-code:
@lfoppiano this is a potential revision that we can discuss next week (I thought it's good to have something to look at).

Advantages:

* can be built from local source and no direct link to github repo
* docker can use cache (although in the current version it is not yet very effective unless we can cache the gradle dependencies, e.g. by mounting the _.gradle_ volume)
* as far as I know, that is how docker is usually used (otherwise the Dockerfile could be outside the repo)
* changes to the project structure / build requirements can be reflected in the Dockerfile together with the code
* you could easily integrate it into the CI build workflow to push docker images for commits

Changes to release:

* rather than building the release image locally, it might be better do that on a CI machine
* otherwise you could instead have a very simple script that checks out a particular version and then builds the docker image (instead of doing the git checkout as part of the docker build)


lfoppiano:
Thanks @de-code. 👍

PR #326, opened 2018-07-05
Update Saxon 9 to a more recent version to remove dependency on saxon9-dom.jar

philgooch:
This removes the downstream dependency on saxon9-dom.jar - see:

https://stackoverflow.com/a/15441957/9098739


kermitt2:
Thanks again !

PR #327, opened 2018-07-05
Assign sanitizedEmails to authorEmailAssigner

philgooch:
Current code sanitizes the emails but then assigns the unsanitized version to authorEmailAssigner. This change ensures that `sanitizedEmails` are correctly assigned.


kermitt2:
thanks a lot @philgooch !

PR #350, opened 2018-10-03
Updated dependencies

lfoppiano:
- dropwizard (to latest version)
 - pdfbox (to latest minor version)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -285,16 +285,16 @@ project(":grobid-service") {
 
     dependencies {
         compile project(':grobid-core')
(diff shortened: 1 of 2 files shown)


lfoppiano:
@kermitt2 IMHO when merging we could squash the commits into a single one, so that we can simplify the history with one clear commit and message. 

FYI Github does that automatically (it works well for simple commits), othwerwise can be done with `rebase`: http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html

PR #367, opened 2019-01-04
[wip] automatically resume model training if model file exists

### LINKED ISSUE #357: Resume model training
(discussed on Mattermost)

This is similar to #178, but rather than encouraging to incrementally learn with only new documents, this is to encourage, or at least enable to resuming the model training with some of the following use cases:

* interrupt model training and resume training later
* resume training with lower error threshold
* resume training with additional documents

Personally I would make it a default as, if this is correct, it would start with a high error if the documents do not match the model well (e.g. many new documents or corpus changed completely). It will load the weights as a head start and then converge towards the optimal model (provided it has all of the documents).

Proposed to:
- resume model training if there is an existing model, start from scratch otherwise
- if resuming training is not desired, the existing model could simply be removed
- alternatively add a flag like `--resume` or `--preload-model <model>`

de-code:
Resolves #357 

This is maybe a bit hacky way.

With that change, it will resume the training of the existing model (still with the whole dataset), if an existing model exists. Otherwise it will start from scratch. There isn't a switch to enable the feature. But one could simply remove the old model to force it to train from scratch.

Some reasons why the current implementation is a bit hacky:

- It copies the old model to the temporary model; Ideally it would just resume from the old model and save to the temp model (but that would probably require changing the signature of the `train` method)
- No tests have been added

/cc @kermitt2


de-code:
Closing this PR. Let me know if you are interested in it.

PR #368, opened 2019-01-04
Use available threads when nb threads is zero

### LINKED ISSUE #358: Use all available cores by default
(discussed on Mattermost)

Currently the number of threads to use can be configured via the `grobid.nb_threads` in the `grobid.properties` file, which by default is `1`.

Proposing to:
- use all available cores if `grobid.nb_threads` is `0` (update comment in `grobid.properties`)
- change default value to `0`

(Patrice mentioned the number of available cores can be determined via `Runtime.getRuntime().availableProcessors();`)

de-code:
Resolves #358 

/cc @kermitt2


kermitt2:
Thanks a lot!

PR #369, opened 2019-01-07
DeLFT integration

kermitt2:
Addition to GROBID of Deep Learning sequence labelling trained with [DeLFT](https://github.com/kermitt2/delft). All neural models are BiLSTM-CRF with Glove embeddings, as alternative to Wapiti CRF. 

Only works with Linux 64 bits for the moment (only 64-bits architectures will be supported). 

To use them: 

- install [DeLFT](https://github.com/kermitt2/delft)
- indicate the path of the DeLFT install in `grobid.properties`
- change the engine from `wapiti` to `delft` in the `grobid-properties` file

Integration is via Java Embedded Python [JEP](https://github.com/ninia/jep), which uses a JNI of CPython. 2 times faster than Tensorflow Java API and significantly faster than RPC serving (see https://www.slideshare.net/FlinkForward/flink-forward-berlin-2017-dongwon-kim-predictive-maintenance-with-apache-flink), does not require to modify DeLFT as it would be the case with Py4J gateway (socket-based). 

There are no neural model for the segmentation and the fulltext models because the input sequence is far too big. The problem would need to be formulated differently for these tasks. 

Low level models not using layout features (author name, dates, affliliations...) perform similarly as CRF, but CRF is much better when layout features are involved (in particular for the header model). Neural models do not use additional features for the moment.

See some evaluation under `grobid-trainer/docs`.

ELMo not yet experimented for the grobid models, but it could make some models better than CRF, although probably too slow for concrete usage (it will make these models 100 times slower in our estimate).

Current neural models are 3-4 time slower than CRF: we do not use batch processing for the moment, not clear how to use batch processing with a cascading approach.


grobid-core/src/main/java/org/grobid/core/jni/JEPThreadPool.java
```diff
@@ -0,0 +1,116 @@
...
+ * we pool the Jep instances in a singleton class. 
+ */
+
+public class JEPThreadPool { 
+    private static final Logger LOGGER = LoggerFactory.getLogger(JEPThreadPool.class);
+
+    private int POOL_SIZE = 1;
+
+    private ExecutorService executor;  
+    private Map<Long,Jep> jepInstances;
```
de-code: Was there a reason not to use [ThreadLocal](https://docs.oracle.com/javase/8/docs/api/java/lang/ThreadLocal.html)?
I guess originally the intention was to use JEP for multiple threads? (judging by the name `pool`)

grobid-core/src/main/java/org/grobid/core/jni/JEPThreadPool.java
```diff
@@ -0,0 +1,116 @@
...
+
+    /**
+     * Hidden constructor
+     */
+    private JEPThreadPool() {
+        // creating a pool of POOL_SIZE threads
+        //executor = Executors.newFixedThreadPool(POOL_SIZE); 
+        executor = Executors.newSingleThreadExecutor();
+        // each of these threads is associated to a JEP instance
+        jepInstances = new HashMap<Long,Jep>();
```
de-code: As far as I am aware, [HashMap](https://docs.oracle.com/javase/8/docs/api/java/util/HashMap.html) is not thread-safe. Not sure if you have synchronisation further up. There is [ConcurrentHashMap](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ConcurrentHashMap.html) if you really need it (without synchronization).

PR #380, opened 2019-02-02
[first iteration] PDFALTO integration

Aazhar:
This is a first iteration to integrate pdfalto.
The main change is replacing former schema with ALTO schema and support character composition(diacritics) from pdfalto.
(reading order and bitmap ocr is not concerned for now)


kermitt2:
@Aazhar I am on it... running some tests this night...

Issue #432, opened 2019-05-16
build.gradle dependency update


GrennanM:
I had a problem with build.gradle. I am using:

- Gradle version: 5.0

- java-8-openjdk

The problem was in relation to the following on line 13:

`classpath 'com.github.jengelman.gradle.plugins:shadow:2.0.2'`

I updated to:

`classpath 'com.github.jengelman.gradle.plugins:shadow:4.0.3'`

The build was then successful.

lfoppiano:
Yes I think we should upgrade to Gradle 5.0 👍 Probably after the next release cause there are too many moving parts that might break 😅

lfoppiano:
I've create a branch gradle5 https://github.com/kermitt2/grobid/tree/gradle5 and the build works. If someone could test more deeply that everything works that would be helpful.

lfoppiano:
I would close this issue, if there are problems, please reopen it

PR #435, opened 2019-06-10
migration to gradle 5

lfoppiano:
Updated to gradle 5 and updated shadow plugin. #432 

I've tested build and running some gradle tasks and it seems to work fine.


kermitt2:
Many thanks Luca !

Issue #409, opened 2019-03-19
JEP support on macOS 


Aazhar:
I have changed the configuration to use delft models but I'm encountering an issue :

> Mar 19, 2019 10:45:56 AM org.grobid.core.main.GrobidHomeFinder getGrobidHomePathOrLoadFromClasspath
WARNING: No Grobid property was provided. Attempting to find Grobid home in the current directory...
Mar 19, 2019 10:45:56 AM org.grobid.core.main.GrobidHomeFinder findGrobidHomeOrFail
WARNING: ***************************************************************
Mar 19, 2019 10:45:56 AM org.grobid.core.main.GrobidHomeFinder findGrobidHomeOrFail
WARNING: *** USING GROBID HOME: /Users/azhar/work/grobid/grobid-home
Mar 19, 2019 10:45:56 AM org.grobid.core.main.GrobidHomeFinder findGrobidHomeOrFail
WARNING: ***************************************************************
Mar 19, 2019 10:45:56 AM org.grobid.core.main.GrobidHomeFinder getGrobidHomePathOrLoadFromClasspath
WARNING: No Grobid property was provided. Attempting to find Grobid home in the current directory...
Mar 19, 2019 10:45:56 AM org.grobid.core.main.GrobidHomeFinder findGrobidHomeOrFail
WARNING: ***************************************************************
Mar 19, 2019 10:45:56 AM org.grobid.core.main.GrobidHomeFinder findGrobidHomeOrFail
WARNING: *** USING GROBID HOME: /Users/azhar/work/grobid/grobid-home
Mar 19, 2019 10:45:56 AM org.grobid.core.main.GrobidHomeFinder findGrobidHomeOrFail
WARNING: ***************************************************************
Mar 19, 2019 10:45:56 AM org.grobid.core.main.GrobidHomeFinder findGrobidPropertiesOrFail
WARNING: Grobid property file location was not explicitly set via 'org.grobid.property' system variable, defaulting to: /Users/azhar/work/grobid/grobid-home/config/grobid.properties
Processing: /Users/azhar/work/grobid/testfiles/kano2012.pdf
Mar 19, 2019 10:45:57 AM org.grobid.core.engines.ProcessEngine inferOutputPath
INFO: No path set for the output directory. Using: /Users/azhar/work/grobid/.
Mar 19, 2019 10:45:58 AM org.grobid.core.main.LibraryLoader load
INFO: Loading external native sequence labelling library
[...]

PR #439, opened 2019-06-21
[wip] fix jep

de-code:
this is work-in-progress...

Trying to get JEP to work locally (Linux).

See #409 (although the issue is related to macOS)

/cc @kermitt2 @lfoppiano

### DIFF
--- a/grobid-core/src/main/java/org/grobid/core/jni/JEPThreadPool.java
+++ b/grobid-core/src/main/java/org/grobid/core/jni/JEPThreadPool.java
@@ -55,7 +55,54 @@ public class JEPThreadPool {
         //executor = Executors.newFixedThreadPool(POOL_SIZE); 
         executor = Executors.newSingleThreadExecutor();
         // each of these threads is associated to a JEP instance
-        jepInstances = new HashMap<>();
+        jepInstances = new ConcurrentHashMap<>();
+    }
+
+    private Jep createJEPInstance() {
+        JepConfig config = new JepConfig();
+        Jep jep = null;
+        boolean success = false;
+        try {
+            File delftPath = new File(GrobidProperties.getInstance().getDeLFTFilePath());
+            if (!delftPath.exists()) {
+                throw new GrobidResourceException("DeLFT installation path does not exist");
+            }
+            if (!delftPath.isDirectory()) {
+                throw new GrobidResourceException("DeLFT installation path is not a directory");
+            }
+            config.addIncludePaths(delftPath.getAbsolutePath());
+            config.setClassLoader(Thread.currentThread().getContextClassLoader());
+            //System.out.println("jep instance thread: " + Thread.currentThread().getId());
+            jep = new Jep(config);
+            jepInstances.put(Thread.currentThread().getId(), jep);
+            // import packages
+            jep.eval("import os");
+            jep.eval("import numpy as np");
+            jep.eval("import keras.backend as K");
+            jep.eval("os.chdir('" + delftPath.getAbsolutePath() + "')");
+            jep.eval("from delft.utilities.Embeddings import Embeddings");
+            jep.eval("import delft.sequenceLabelling");
+            jep.eval("from delft.sequenceLabelling import Sequence");
+            jep.eval("from delft.sequenceLabelling.reader import load_data_and_labels_crf_file");
+            jep.eval("from delft.sequenceLabelling.reader import load_data_crf_string");
+            jep.eval("from sklearn.model_selection import train_test_split");
+            success = true;
+            return jep;
+        } catch(JepException e) {
+            LOGGER.error("JEP initialization failed", e);
+            throw new RuntimeException("JEP initialization failed", e);
+        } catch(GrobidResourceException e) {
+            LOGGER.error("DeLFT installation path invalid, JEP initialization failed", e);
+            throw new RuntimeException("DeLFT installation path invalid, JEP initialization failed", e);
+        } finally {
+            if (!success) {
+                try {
+                    jep.close();
+                } catch (JepException e) {
+                    LOGGER.error("failed to close JEP instance", e);
+                }
+            }
+        }
     }
 
     /**
@@ -64,40 +111,20 @@ public class JEPThreadPool {
      * (or create one the first time). 
      */
     public Jep getJEPInstance() {
-        if (jepInstances.get(Thread.currentThread().getId()) == null) {
-            JepConfig config = new JepConfig();
-            try {
-                File delftPath = new File(GrobidProperties.getInstance().getDeLFTFilePath());
-                if (!delftPath.exists()) {
-                    throw new GrobidResourceException("DeLFT installation path does not exist");
-                }
-                if (!delftPath.isDirectory()) {
-                    throw new GrobidResourceException("DeLFT installation path is not a directory");
-                }
-                config.addIncludePaths(delftPath.getAbsolutePath());
-                config.setClassLoader(Thread.currentThread().getContextClassLoader());
-                //System.out.println("jep instance thread: " + Thread.currentThread().getId());
-                try(Jep jep = new Jep(config)) {
-                    jepInstances.put(Thread.currentThread().getId(), jep);
-                    // import packages
-                    jep.eval("import os");
-                    jep.eval("import numpy as np");
-                    jep.eval("import keras.backend as K");
-                    jep.eval("os.chdir('" + delftPath.getAbsolutePath() + "')");
-                    jep.eval("from delft.utilities.Embeddings import Embeddings");
-                    jep.eval("import delft.sequenceLabelling");
-                    jep.eval("from delft.sequenceLabelling import Sequence");
-                    jep.eval("from delft.sequenceLabelling.reader import load_data_and_labels_crf_file");
-                    jep.eval("from delft.sequenceLabelling.reader import load_data_crf_string");
-                    jep.eval("from sklearn.model_selection import train_test_split");
-                }
-            } catch(JepException e) {
-                LOGGER.error("JEP initialization failed", e);
-            } catch(GrobidResourceException e) {
-                LOGGER.error("DeLFT installation path invalid, JEP initialization failed", e);
-            }
+        long threadId = Thread.currentThread().getId();
+        Jep jep = jepInstances.get(threadId);
+        if (jep == null) {
+            jep = this.createJEPInstance();
+            jepInstances.put(threadId, jep);
         }
-        return jepInstances.get(Thread.currentThread().getId());
+        try {
+            jep.isValidThread();
+        } catch (JepException e) {
+            LOGGER.warn("JEP instance no longer usable, creating new instance", e);
+            jep = this.createJEPInstance();
+            jepInstances.put(threadId, jep);
+        }
+        return jep;
     }
 
     public void run(Runnable task) throws InterruptedException {
--- a/grobid-core/src/main/java/org/grobid/core/jni/JEPThreadPool.java
+++ b/grobid-core/src/main/java/org/grobid/core/jni/JEPThreadPool.java
@@ -114,6 +114,7 @@ public class JEPThreadPool {
         long threadId = Thread.currentThread().getId();
         Jep jep = jepInstances.get(threadId);
         if (jep == null) {
+            LOGGER.info("Creating JEP instance for thread " + threadId);
             jep = this.createJEPInstance();
             jepInstances.put(threadId, jep);
         }
--- a/grobid-core/src/main/java/org/grobid/core/jni/JEPThreadPool.java
+++ b/grobid-core/src/main/java/org/grobid/core/jni/JEPThreadPool.java
@@ -100,7 +100,7 @@ public class JEPThreadPool {
             File delftPath = this.getAndValidateDelftPath();
             JepConfig config = this.getJepConfig(
                 delftPath,
-                PythonVirtualEnvConfig.getInstance().getSitePackagesPath()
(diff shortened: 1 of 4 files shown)


grobid-core/src/main/java/org/grobid/core/main/LibraryLoader.java
```diff
@@ -158,7 +158,9 @@ public boolean accept(File dir, String name) {
                     if (StringUtils.isNotEmpty(GrobidProperties.getPythonVirtualEnv())) {
                         String virtualEnv = GrobidProperties.getPythonVirtualEnv() + File.separator + "lib";
 
-                        List<Path> pythons = Files.find(Paths.get(virtualEnv), 1, (path, attr) -> path.getFileName().toString().startsWith("libpython3.6m")).collect(Collectors.toList());
+                        List<Path> pythons = Files.find(Paths.get(virtualEnv), 1, (path, attr) -> path.getFileName().toString().startsWith(
+                            "python3.6"
```
de-code: this needs to be more flexible, maybe remove filtering?

grobid-core/src/main/java/org/grobid/core/main/LibraryLoader.java
```diff
@@ -168,11 +170,13 @@ public boolean accept(File dir, String name) {
...
                         if (CollectionUtils.isEmpty(pythonVersions)) {
-                            throw new GrobidException("Cannot find a suitable version (3.5 or 3.6) of python in your virtual environment. ");
+                            throw new GrobidException(
+                                "Cannot find a suitable version (3.5 or 3.6) of python in your virtual environment: " +
+                                virtualEnv
+                            );
                         }
 
                         addLibraryPath(virtualEnv);
-                        System.loadLibrary("python" + pythonVersions.get(0));
```
de-code: why did we need that?

grobid-core/src/main/java/org/grobid/core/jni/JEPThreadPool.java
```diff
@@ -55,7 +55,54 @@ private JEPThreadPool() {
...
+            jep.eval("import delft.sequenceLabelling");
+            jep.eval("from delft.sequenceLabelling import Sequence");
+            jep.eval("from delft.sequenceLabelling.reader import load_data_and_labels_crf_file");
+            jep.eval("from delft.sequenceLabelling.reader import load_data_crf_string");
+            jep.eval("from sklearn.model_selection import train_test_split");
+            success = true;
+            return jep;
+        } catch(JepException e) {
+            LOGGER.error("JEP initialization failed", e);
+            throw new RuntimeException("JEP initialization failed", e);
```
de-code: this was returning `null` before (after logging the error). Probably best if it throws an exception.

grobid-core/src/main/java/org/grobid/core/jni/JEPThreadPool.java
```diff
@@ -64,40 +111,20 @@ private JEPThreadPool() {
...
-                LOGGER.error("DeLFT installation path invalid, JEP initialization failed", e);
-            }
+        long threadId = Thread.currentThread().getId();
+        Jep jep = jepInstances.get(threadId);
+        if (jep == null) {
+            jep = this.createJEPInstance();
+            jepInstances.put(threadId, jep);
+        }
+        try {
+            jep.isValidThread();
```
de-code: This is not strictly necessary. It's called whenever you call `jep.eval`. It just makes it fail fast or re-create the instance. Although it shouldn't randomly close.

grobid-core/src/main/java/org/grobid/core/jni/JEPThreadPool.java
```diff
@@ -64,40 +111,20 @@ private JEPThreadPool() {
...
-                if (!delftPath.exists()) {
-                    throw new GrobidResourceException("DeLFT installation path does not exist");
-                }
-                if (!delftPath.isDirectory()) {
-                    throw new GrobidResourceException("DeLFT installation path is not a directory");
-                }
-                config.addIncludePaths(delftPath.getAbsolutePath());
-                config.setClassLoader(Thread.currentThread().getContextClassLoader());
-                //System.out.println("jep instance thread: " + Thread.currentThread().getId());
-                try(Jep jep = new Jep(config)) {
```
de-code: This was closing the `jep` instance after the initialisation.

grobid-core/src/main/java/org/grobid/core/jni/JEPThreadPool.java
```diff
@@ -55,7 +55,54 @@ private JEPThreadPool() {
         //executor = Executors.newFixedThreadPool(POOL_SIZE); 
         executor = Executors.newSingleThreadExecutor();
         // each of these threads is associated to a JEP instance
-        jepInstances = new HashMap<>();
+        jepInstances = new ConcurrentHashMap<>();
```
de-code: [HashMap](https://docs.oracle.com/javase/8/docs/api/java/util/HashMap.html) is not thread-safe. Using [ConcurrentHashMap](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ConcurrentHashMap.html) instead. Although I would recommend using [ThreadLocal](https://docs.oracle.com/javase/8/docs/api/java/lang/ThreadLocal.html). (See [comment](https://github.com/kermitt2/grobid/pull/369/files#r296194872), [comment](https://github.com/kermitt2/grobid/pull/369/files#r296184959))

de-code:
It might be best not to ship a JEP binary with GROBID as it seems to tie it to closely to the Python version.

kermitt2:
> It might be best not to ship a JEP binary with GROBID as it seems to tie it too closely to the Python version.

Actually it's the idea to have everything closely tied :)

To have everything working, the user will have to install DeLFF, which means that python 3.5 lib are available in her/his system (not just in an environment). The JEP binary shipped with GROBID is depending on python 3.5, but it will work because we are sure that it's installed and available. 

If it would not be possible to ship JEP binaries with GROBID that work after DeLFT is installed, then all the building process would be very complicated for the user.

grobid-core/src/main/java/org/grobid/core/main/LibraryLoader.java
kermitt2: DeLFT is tested with python 3.5, I have not considered 3.6 when integrating with GROBID. To support different version of python, we would need to ship different version of JEP binaries and adapt the library path to be added based on the python version used to build DeLFT.  

This filtering part has been written by Luca, I've not looked at it yet!

grobid-core/src/main/java/org/grobid/core/main/LibraryLoader.java
kermitt2: I am lost in the branches :D

grobid-core/src/main/java/org/grobid/core/jni/JEPThreadPool.java
kermitt2: yes, I have not reviewed that branch

grobid-core/src/main/java/org/grobid/core/jni/JEPThreadPool.java
kermitt2: yes very good point ! This might cause errors when 2 threads try to allocate concurrently JEP instances.

de-code:
This currently works for me provided I activate the Python environment and point LD_LIBRARY_PATH to the `jep` package directory.

```bash
LD_LIBRARY_PATH=/path/to/delft/venv/lib/python3.6/site-packages/jep \
  ./gradlew run --no-daemon
```

Need to see how we could avoid specifying `LD_LIBRARY_PATH`. It should also not need to activate the environment with `grobid.delft.python.virtualEnv` being specified. On the other hand it would also be good not having to specify `grobid.delft.python.virtualEnv` but instead use the activated environment.

grobid-core/src/main/java/org/grobid/core/jni/JEPThreadPool.java
kermitt2: I think there might be cases where the python part crash the JEP instance. Although I don't remember exactly why I put this, maybe just paranoia.

de-code:
> Actually it's the idea to have everything closely tied :)
> 
> To have everything working, the user will have to install DeLFF, which means that python 3.5 lib are available in her/his system (not just in an environment). The JEP binary shipped with GROBID is depending on python 3.5, but it will work because we are sure that it's installed and available.

Not many people have that many Python versions installed. Sometimes it's more effort trying to upgrade / downgrade, maybe requiring a package outside the package manager. For example I am on Ubuntu 18.04 (LTS) without Python 3.5 being available. (It's different when using Docker of course)

> If it would not be possible to ship JEP binaries with GROBID that work after DeLFT is installed, then all the building process would be very complicated for the user.

By user, do you mean the end-user? Would they not use Docker?
Otherwise I am currently not seeing how it's more difficult. They already need to have the DeLFT dependencies installed? Why can't JEP be one of the dependencies in that case?

grobid-core/src/main/java/org/grobid/core/main/LibraryLoader.java
de-code: I have changed that to _contains_ `python3` for now.

kermitt2:
> Not many people have that many Python versions installed. 

Indeed it's an issue, we could build and ship JEP for different mainstream Python versions. It seems from your previous comment that same JEP binary would work with 3.5 and 3.6. 

> By user, do you mean the end-user? Would they not use Docker?

I think GROBID is a tool for developer, integrator, etc. they are the end-users, and might not use the GROBID docker in majority (apart for discovering the tool and deployment), but I might be wrong of course. 

> Otherwise I am currently not seeing how it's more difficult. They already need to have the DeLFT dependencies installed? Why can't JEP be one of the dependencies in that case?

They would need to rebuild the JEP JNI binaries, no? Normally installing DeLFT is just 2 commands that we can't avoid, then same for GROBID, and I think it's helpful if they don't have to build other dependencies.  

As alternative to JEP, if it is finally too complicated to have something working "out-of-the-box" with it, we could use a RPC/REST serving or socket approach for DeLFT (and put all DeLFT gulash in a docker why not). The main motivation for JEP is that it's the fastest python integration framework to Java available.

de-code:
> Indeed it's an issue, we could build and ship JEP for different mainstream Python versions. It seems from your previous comment that same JEP binary would work with 3.5 and 3.6.

No, it didn't work with the shipped binary. Instead I installed `jep` in the virtual environment and pointed LD_LIBRARY_PATH to it. I don't think it is using the shipped binary in that case, or is it? (I don't really know what it loads)

> They would need to rebuild the JEP JNI binaries, no? Normally installing DeLFT is just 2 commands that we can't avoid, then same for GROBID, and I think it's helpful if they don't have to build other dependencies.

I installed `jep` via `pip install jep`. I found that easier than many other steps :D

> As alternative to JEP, if it is finally too complicated to have something working "out-of-the-box" with it, we could use a RPC/REST serving or socket approach for DeLFT (and put all DeLFT gulash in a docker why not). The main motivation for JEP is that it's the fastest python integration framework to Java available.

The working  "ouf-of-the-box" seems quite important to me. I would prioritise it over speed. Maybe there could be two implementations if necessary. I wouldn't give up on JEP for "out-of-the-box" just yet.

The idealised Docker setup would decouple everything via microservices. In a way having the REST service would be good as a clear separation and I could replace that more easily, especially for experimentation. That being said, seeing DeLFT as a very close cousin, it would seem reasonable to me to offer a Docker container that contains both GROBID and DeLFT - in fact Wapiti is currently also included in the same Docker container. I would probably offer two different images / tags (for reduced size).

de-code:
BTW when I try to run it without having activated the virtual environment before, I am getting:

```
ERROR [2019-06-21 15:06:17,316] org.grobid.core.jni.JEPThreadPool: JEP initialization failed
! jep.JepException: JEP THREAD WARNING: Unsafe reuse of thread pool-4-thread-1 for another Python sub-interpreter.
! Please close() the previous Jep instance to ensure stability.
! at jep.Jep.<init>(Jep.java:239)
! at jep.Jep.<init>(Jep.java:228)
```

de-code:
> BTW when I try to run it without having activated the virtual environment before, I am getting:
> 
> ```
> ERROR [2019-06-21 15:06:17,316] org.grobid.core.jni.JEPThreadPool: JEP initialization failed
> ! jep.JepException: JEP THREAD WARNING: Unsafe reuse of thread pool-4-thread-1 for another Python sub-interpreter.
> ! Please close() the previous Jep instance to ensure stability.
> ! at jep.Jep.<init>(Jep.java:239)
> ! at jep.Jep.<init>(Jep.java:228)
> ```

Actually I missed a previous error that explains it:

```
ERROR [2019-06-21 15:38:07,600] org.grobid.core.jni.JEPThreadPool: JEP initialization failed
! jep.JepException: <class 'ModuleNotFoundError'>: No module named 'jep'
! at <string>.<module>(<string>:1)
! at jep.Jep.eval(Native Method)
```

de-code:
I didn't really get it to work without activating the virtual environment first. It seems it requires more hackery and I am not sure whether there is much benefit. Why not just activate the virtual environment before starting GROBID?

It will now automatically add the `jep` directory in site-packages to the library path (assuming jep is installed via pip). In that case it is definitely not loading the jep library in `lin-64` (I removed it locally).

I wonder whether that also works with a `conda` environment or on a mac.

lfoppiano:
@de-code my code was based on mac and on conda, I will have a look next week.

lfoppiano:
I've merged, and fixed mac osx. I've also add a brutal workaround for to avoid failing when no virtualenv is specified: now is just using the one specified in the configuration.

PR #446, opened 2019-06-27
use https:// in generated links (not http://)

### LINKED ISSUE #444: https:// links for doi.org and arxiv.org
For general internet use the consensus is that HTTPS is better practice than HTTP.

GROBID generates links in TEI-XML in a couple places. I made the change from `http://` to `https://` in two places in this patch: https://github.com/bnewbold/grobid/commit/51edc9db01b1278932dd0d1ab547bb072e12e675

If you think this would be an improvement I can submit a PR after fixing any tests and hunting down any other instances (eg, in the PDF overlay code path?).

bnewbold:
Closes #444


kermitt2:
Thank you Bryan

PR #448, opened 2019-06-27
revise jep v2

de-code:
further revisions... see comments below...

### DIFF
--- a/grobid-core/src/main/java/org/grobid/core/jni/JEPThreadPool.java
+++ b/grobid-core/src/main/java/org/grobid/core/jni/JEPThreadPool.java
@@ -128,7 +128,7 @@ public class JEPThreadPool {
      * The method will return to the thread its dedicated Jep instance
      * (or create one the first time).
      */
-    public Jep getJEPInstance() {
+    public synchronized Jep getJEPInstance() {
         long threadId = Thread.currentThread().getId();
         Jep jep = jepInstances.get(threadId);
         if (jep == null) {
--- a/grobid-core/src/main/java/org/grobid/core/jni/PythonEnvironmentConfig.java
+++ b/grobid-core/src/main/java/org/grobid/core/jni/PythonEnvironmentConfig.java
@@ -77,8 +77,6 @@ public class PythonEnvironmentConfig {
         }
         if (StringUtils.isEmpty(virtualEnv)) {
             virtualEnv = activeVirtualEnv;
-        } else if(StringUtils.isEmpty(activeVirtualEnv)) {
-            activeVirtualEnv = virtualEnv;
         }
 
         List<Path> pythons;
@@ -115,7 +113,7 @@ public class PythonEnvironmentConfig {
             Paths.get(virtualEnv),
             sitePackagesPath,
             jepPath,
-            activeVirtualEnv != null && virtualEnv.equals(activeVirtualEnv)
+            StringUtils.equals(virtualEnv, activeVirtualEnv)
         );
     }
 
--- a/grobid-core/src/main/java/org/grobid/core/jni/PythonEnvironmentConfig.java
+++ b/grobid-core/src/main/java/org/grobid/core/jni/PythonEnvironmentConfig.java
@@ -44,6 +44,13 @@ public class PythonEnvironmentConfig {
         return this.sitePackagesPath;
     }
 
+    public Path getNativeLibPath() {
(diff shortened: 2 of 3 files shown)


grobid-core/src/main/java/org/grobid/core/jni/JEPThreadPool.java
```diff
@@ -128,7 +128,7 @@ private Jep createJEPInstance() {
      * The method will return to the thread its dedicated Jep instance
      * (or create one the first time).
      */
-    public Jep getJEPInstance() {
+    public synchronized Jep getJEPInstance() {
```
de-code: I realised it should really be synchronized. Otherwise it could try to initialise two instances at the same time. One could use concurrent checks in a way that doesn't require synchronization. But not sure if it's worth it.

grobid-core/src/main/java/org/grobid/core/main/LibraryLoader.java
```diff
@@ -192,8 +191,9 @@ public static void addLibraryPath(String pathToAdd) throws Exception {
             if (path.equals(pathToAdd))
                 return;
 
-        String[] newPaths = Arrays.copyOf(paths, paths.length + 1);
-        newPaths[newPaths.length - 1] = pathToAdd;
+        String[] newPaths = new String[paths.length + 1];
```
de-code: adding libraries to the front instead. as an added effect it will then prefer `jep` from the virtual env rather than the one in grobid-home.

grobid-core/src/main/java/org/grobid/core/jni/PythonEnvironmentConfig.java
```diff
@@ -99,10 +119,18 @@ public static PythonEnvironmentConfig getInstanceForVirtualEnv(String virtualEnv
         );
     }
 
+    public static String getActiveVirtualEnv() {
+        String activeVirtualEnv = System.getenv("VIRTUAL_ENV");
+        if (StringUtils.isEmpty(activeVirtualEnv)) {
+            activeVirtualEnv = System.getenv("CONDA_PREFIX");
```
de-code: now also detecting an activated conda virtual env

grobid-core/src/main/java/org/grobid/core/jni/PythonEnvironmentConfig.java
```diff
@@ -69,7 +86,10 @@ public static PythonEnvironmentConfig getInstanceForVirtualEnv(String virtualEnv
             pythons = Files.find(
                 Paths.get(virtualEnv, "lib"),
                 1,
-                (path, attr) -> path.getFileName().toString().contains("python3")
+                (path, attr) -> (
+                    path.toFile().isDirectory()
```
de-code: In conda the lib folder is full of stuff and I can see now why we need the filtering. Just looking for directories seem to make sense (it was picking up the wrong one otherwise).

grobid-core/src/main/java/org/grobid/core/jni/PythonEnvironmentConfig.java
```diff
@@ -44,6 +44,23 @@ public Path getSitePackagesPath() {
         return this.sitePackagesPath;
     }
 
+    public Path getNativeLibPath() {
+        if (this.virtualEnv == null) {
+            return null;
+        }
+        return Paths.get(this.virtualEnv.toString(), "lib");
```
de-code: This is really only useful for conda virtual envs. But it doesn't really matter if we add a directory without libraries.

grobid-core/src/main/java/org/grobid/core/jni/PythonEnvironmentConfig.java
```diff
@@ -60,16 +77,17 @@ public static PythonEnvironmentConfig getInstanceForVirtualEnv(String virtualEnv
         }
         if (StringUtils.isEmpty(virtualEnv)) {
             virtualEnv = activeVirtualEnv;
-        } else if(StringUtils.isEmpty(activeVirtualEnv)) {
```
de-code: This didn't seem to make sense to me. What wasn't working without setting `activeVirtualEnv`? It seems `isActive` isn't actually used, therefore I am not sure what effect it should have.

lfoppiano: no problem with that, let's remove it

PR #452, opened 2019-06-28
revise jep v3

de-code:
This is what got lost in the merge

/cc @lfoppiano

### DIFF
--- a/grobid-core/src/main/java/org/grobid/core/main/LibraryLoader.java
+++ b/grobid-core/src/main/java/org/grobid/core/main/LibraryLoader.java
@@ -149,26 +149,21 @@ public class LibraryLoader {
                 // java.library.path (JEP will anyway try to load from java.library.path, so explicit file 
                 // loading here will not help)
                 try {
+                    addLibraryPath(libraryFolder.getAbsolutePath());
 
                     PythonEnvironmentConfig pythonEnvironmentConfig = PythonEnvironmentConfig.getInstance();
                     if (pythonEnvironmentConfig.isEmpty()) {
                         LOGGER.info("no python environment configured");
                     } else {
                         LOGGER.info("configuring python environment: " + pythonEnvironmentConfig.getVirtualEnv());
-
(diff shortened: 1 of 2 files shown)


grobid-core/src/main/java/org/grobid/core/main/LibraryLoader.java
```diff
@@ -149,26 +149,21 @@ public boolean accept(File dir, String name) {
...
-                            addLibraryPath(libraryFolder.getAbsolutePath());
-                            LOGGER.info("adding library path " + pythonEnvironmentConfig.getJepPath());
-                            addLibraryPath(pythonEnvironmentConfig.getJepPath().toString());
-//                        System.setProperty("java.library.path", System.getProperty("java.library.path") + ":" + LibraryLoader.getLibraryFolder());
-
-                            addLibraryPath(pythonEnvironmentConfig.getNativeLibPath().toString());
                             System.loadLibrary("python" + pythonEnvironmentConfig.getPythonVersion() + "m");
                             System.loadLibrary(DELFT_NATIVE_LIB_NAME);
                         } else if (SystemUtils.IS_OS_LINUX) {
-                            System.setProperty("java.library.path", System.getProperty("java.library.path") + ":" + pythonEnvironmentConfig.getVirtualEnv() + File.separator + "lib");
```
de-code: I am not actually sure whether we need to update `java.library.path`. It doesn't seem to have an effect. If we do, we probably want to encapsulate updating both in a single function, e.g. within `addLibraryPath`.

grobid-core/src/main/java/org/grobid/core/main/LibraryLoader.java
```diff
@@ -149,26 +149,21 @@ public boolean accept(File dir, String name) {
                 // java.library.path (JEP will anyway try to load from java.library.path, so explicit file 
                 // loading here will not help)
                 try {
+                    addLibraryPath(libraryFolder.getAbsolutePath());
```
de-code: We should now add libraryFolder before adding `getNativeLibPaths`..

PR #457, opened 2019-07-07
[wip] Coordinates output in TEI

lfoppiano:
This implementation was done to include coordinates into the TEI output. It requires testing. 

~Fulltext and~ Header parser ~were~ was updated using the `TaggingTokenClusteror`.


kermitt2:
Thanks you, that's cool, I have almost forgotten that work ! 

I'll review it, it will probably require quite a lot of tests because the way the interrupted field in the header was concatenated in the previous version was hacky (with a " ; " usually, which is exploited at some place to known that the field was distributed in separated place of the header) and this is a good time to rework that. The current concatenation here in the PR will make those post-processing (in particular for affiliation attachment) to fail.

lfoppiano:
The parameters `TeiCoordinates` should be renamed `outputCoordinates`, so that it's independent from the format. I'm not sure it's better to repeat the parameters rather than using just one...

kermitt2:
Actually for the parameter `teiCoordinates`, it is tight by choice to the TEI and to the name of the TEI structure to be enriched with coordinates. I think it's much easier, we just look at the resulting TEI and select the element name where we want coordinates. Otherwise, it introduces an additional layer of naming, mapping, abstraction, etc. which will be confusing and don't really help. 

The resulting TEI can then always be converted into another format and the coordinates too, in case the alternate target format supports coordinates. 

More generally, in the global design of GROBID, the document model is TEI because it is very general and generic, and I think we don't want to add complexity with document model independence.

lfoppiano:
> Actually for the parameter `teiCoordinates`, it is tight by choice to the TEI and to the name of the TEI structure to be enriched with coordinates. I think it's much easier, we just look at the resulting TEI and select the element name where we want coordinates. Otherwise, it introduces an additional layer of naming, mapping, abstraction, etc. which will be confusing and don't really help.
> 
> The resulting TEI can then always be converted into another format and the coordinates too, in case the alternate target format supports coordinates.
> 
> More generally, in the global design of GROBID, the document model is TEI because it is very general and generic, and I think we don't want to add complexity with document model independence.

No problem, it was a suggestion. I understand and agree with you.

lfoppiano:
it seems this is already implemented, isn't it? 

@kermitt2 feel free to close it

kermitt2:
Yes HeaderParser was re-written using TaggingTokenClusteror, removing the hacky aspects and covering author coordinates.

PR #459, opened 2019-07-15
add support for delft layout features for label task

de-code:
This adds support for DeLFT layout features on the GROBID side.
To actually work, it would also need the support on the DeLFT side (which I have currently implemented in [sciencebeam-trainer-delft](https://github.com/elifesciences/sciencebeam-trainer-delft)).
It should be backward compatible.

I added an optional configuration option for the package, which allows me to use my own Python package rather than `delft` (which I use in part, but currently extend heavily - until a similar change is ported back)

This is currently only implemented for the label task (not the train task).

Not sure whether implementing it for the training task is a priority. Judging by the comments, the JNI method for training doesn't seem to be stable. And the `grobidTagger.py` of the current DeLFT version itself doesn't support it anyway.

/cc @kermitt2


lfoppiano:
I believe this was implemented at some point. I'm closing this PR as it's quite old.

PR #463, opened 2019-07-25
Add a label for line numbers to the segmentation model

Vitaliy-1:
Hi @kermitt2,

Can you take a look at this PR? Am I understanding your recommendations that we have discussed earlier regarding line numbers right? Thanks!


Vitaliy-1:
Probably forgot to modify TEIFormatter...

Vitaliy-1:
On second thought, I doubt that anyone will want to expose line numbers in the resulted TEI, so they should be probably removed from TEIFormater. Am I understanding this process right?

Vitaliy-1:
Thanks!

kermitt2:
Hello @Vitaliy-1 !

Given that line numbers are something to ignore in term of structure, they are just some presentation related artifacts valid for one particular layout, it's not clear that introducing a label for this is useful. 

Currently in the segmentation model, everything "ignored" stays outside any tags. It simplifies the manual annotation and makes the model less complex (because there are less labels) and faster to train. I am not sure about the impact on accuracy (it was very good with citations), it might depend on what kind of "noise" is considered.  

The main issue I think with line numbers is that they are recognized well as one full block by pdfalto. Sometimes they appear as one line number per line in the text body/header block which is the issue. Ideally we would like to have the line number recognized as one independent column, that will be filtered out by the segmentation model. If we start annotating line numbers in the segmentation model before tackling the issue in pdfalto, the problem is that these annotations might need to be redone with a different reading order - in particular with an explicit label for this. 

What do you think?

Vitaliy-1:
This makes sense. If it's something where PDFAlto can be improved, line numbers can be just filtered out by the model.

I'm preparing annotated data for segmentation model and will keep those introduced tags for now.

kermitt2:
As the support for line numbers has been added via low level recognition in pdfalto, we don't need corresponding labels in grobid.

PR #469, opened 2019-07-26
custom delft train args

de-code:
Allows custom args to be passed in to the the training operation.

It adds the following optional configuration options:
- `grobid.delft.train.module`: Training module (default is `grobidTagger.py` as per `DeLFTModel`)
- `grobid.delft.train.args`: additional arguments to be passed to the training module. To keep it simple, it's split on spaces (which could be improved, should shell parsing be required in the future).

This is also useful for the current training, to configure the training.

The `useELMo` flag didn't actually work for training because the `command` list was immutable.

/cc @kermitt2 @lfoppiano

### DIFF
--- a/grobid-core/src/main/java/org/grobid/core/jni/DeLFTModel.java
+++ b/grobid-core/src/main/java/org/grobid/core/jni/DeLFTModel.java
@@ -211,12 +211,13 @@ public class DeLFTModel {
     }
 
     protected static List<String> getTrainCommand(String modelName, File trainingData) {
-        List<String> command = Arrays.asList("python3", 
+        List<String> command = new ArrayList<>(Arrays.asList("python3", 
             "grobidTagger.py", 
             modelName,
             "train",
             "--input", trainingData.getAbsolutePath(),
-            "--output", GrobidProperties.getModelPath().getAbsolutePath());
+            "--output", GrobidProperties.getModelPath().getAbsolutePath()
+        ));
         if (GrobidProperties.useELMo()) {
             command.add("--use-ELMo");
         }
--- a/grobid-core/src/main/java/org/grobid/core/jni/DeLFTModel.java
+++ b/grobid-core/src/main/java/org/grobid/core/jni/DeLFTModel.java
@@ -210,9 +210,7 @@ public class DeLFTModel {
         } 
     }
 
-    protected static List<String> getTrainCommand(
-        String modelName, File trainingData, File outputModel
-    ) {
+    protected static List<String> getTrainCommand(String modelName, File trainingData) {
         List<String> command = Arrays.asList("python3", 
             "grobidTagger.py", 
             modelName,
@@ -232,7 +230,7 @@ public class DeLFTModel {
     public static void train(String modelName, File trainingData, File outputModel) {
         try {
             LOGGER.info("Train DeLFT model " + modelName + "...");
-            List<String> command = getTrainCommand(modelName, trainingData, outputModel);
+            List<String> command = getTrainCommand(modelName, trainingData);
             LOGGER.info("Running: {}", command);
 
             ProcessBuilder pb = new ProcessBuilder(command);
--- a/grobid-core/src/main/java/org/grobid/core/jni/DeLFTModel.java
+++ b/grobid-core/src/main/java/org/grobid/core/jni/DeLFTModel.java
@@ -213,8 +213,12 @@ public class DeLFTModel {
     }
 
     protected static List<String> getTrainCommand(String modelName, File trainingData) {
+        String trainModule = GrobidProperties.getDeLFTTrainModule();
+        if (StringUtils.isEmpty(trainModule)) {
+            trainModule = "grobidTagger.py";
+        }
         List<String> command = new ArrayList<>(Arrays.asList("python3", 
-            "grobidTagger.py", 
+            trainModule,
             modelName,
             "train",
             "--input", trainingData.getAbsolutePath(),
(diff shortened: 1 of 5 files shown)


lfoppiano:
@de-code so this should run the training using delft directly from Grobid?

de-code:
> @de-code so this should run the training using delft directly from Grobid?

Yes, it would be trained like the Wapiti model. Although I have deferred trying to run it that way. Not least because I then need to create GPU version of the Docker container to run in the cloud. Since I am just using a single training dataset at present, it's okay to do that manually (I still need to run the GROBID training command to get the converted training data).

grobid-core/src/main/java/org/grobid/core/jni/DeLFTModel.java
```diff
@@ -208,7 +210,30 @@ public void run() {
                 LOGGER.error("DeLFT model training via JEP failed", e);
             } 
         } 
-    } 
+    }
+
+    protected static List<String> getTrainCommand(String modelName, File trainingData) {
+        String trainModule = GrobidProperties.getDeLFTTrainModule();
```
lfoppiano: I think here you should export the ENV variable with the conda environment... or check if it exists and use it when running the command... 

Right now, running it from a submodule, I get:

```
18:18:35.610 [main] INFO org.grobid.core.jni.DeLFTModel - Train DeLFT model superconductors...
18:18:35.610 [main] INFO org.grobid.core.jni.DeLFTModel - Running: [python3, grobidTagger.py, superconductors, train, --input, /Users/lfoppiano/development/projects/grobid/grobid-home/tmp/superconductors3567936683595801467.train, --output, /Users/lfoppiano/development/projects/grobid/grobid-superconductors/../grobid-home/models]
Traceback (most recent call last):
  File "grobidTagger.py", line 3, in <module>
    from delft.utilities.Embeddings import Embeddings
  File "/Users/lfoppiano/development/projects/delft/delft/utilities/Embeddings.py", line 3, in <module>
    from keras.preprocessing import text, sequence
ModuleNotFoundError: No module named 'keras'
```

lfoppiano: Any though how to fix this? I'm not sure what we can do because by running the command imply we need to load the whole environment

de-code: Have you tried activating the environment first? I would argue it's outside the scope of this PR.
But as with using JEP in general, personally I would leave activating the environment to the user as trying to activate it programatically with JEP seems less stable.

lfoppiano: OK sorry if I run it after having activated the environment, it works. 
It doesn't work, unfortunately, using the CONDA env variable, because is not just a matter of libraries but also paths... since the training is ran using python command 😢 

No idea if this can be fixed...

de-code: > OK sorry if I run it after having activated the environment, it works.
> It doesn't work, unfortunately, using the CONDA env variable, because is not just a matter of libraries but also paths... since the training is ran using python command 
> 
> No idea if this can be fixed...

Not sure if I understand you correctly. Everything works when activating the conda environment first?

Not sure about the CONDA env variable, do you mean `CONDA_PREFIX`?. I guess that is related to the Python Environment config and the virtual env config in `grobid.properties` that you added? When running an external command, it is certainly easier to properly _activate_ a virtual environment as environment variables can be passed in (without hacking some Java internals). You could just add `${CONDA_PREFIX}/bin` to (the beginning of) the `PATH`.

lfoppiano: I wonder if we could read the CONDA_PREFIX and pick up python from that location instead of just calling a generic `python3` executable... 

In this way we maintain the double compatibility: 
 1. environment activation + command launch and  
 2. environment setting + command launch 

Luca

de-code: I've created a separate issue for that

lfoppiano:
This fix works in Grobid an on a submodule 👍 the training must be ran after activating the virtual environment. 
Better to fix #490 on a separate PR, cause it requires moving around some code.

de-code:
Hi @lfoppiano I rebased as requested

de-code:
BTW I would recommend to always squash merge PRs

lfoppiano:
I wonder, what about we make this an optional method for training the delft model? 
Since Jep is not reliable at this stage, this could still be removed in future. 

For me, personally the alternative is to poke the files from the tmp directory, and run the training manually, which is also not ideal  dd

de-code:
> I wonder, what about we make this an optional method for training the delft model?
> Since Jep is not reliable at this stage, this could still be removed in future.
> 
> For me, personally the alternative is to poke the files from the tmp directory, and run the training manually, which is also not ideal dd

Just trying to go through some old GitHub notifications. Not sure if it was for me but there seem to be [some agreement, that a way to train a DeLFT model without GROBID can be beneficial](https://github.com/kermitt2/delft/issues/42#issuecomment-568765771).

For training via GROBID, as you mentioned JEP is less reliable. Maybe training via the command has the additional advantage that it is more transparent, simpler (given the tool already exists) and could be reproduced separately (via the CLI). I guess for training and evaluation there shouldn't be any noticeable performance difference.

lfoppiano:
I would parametrize this change via configuration and add a new CLI function in grobid to generate the training data automatically so that we provide multiple ways of training the models.

lfoppiano:
new CLI is done: https://github.com/lfoppiano/grobid-superconductors/blob/master/src/main/java/org/grobid/service/command/PrepareDelftTraining.java 

Since it's in a (yet) private repository  https://gist.github.com/lfoppiano/dd28365a83d7aaf9b459ec0ef846ea34

de-code:
I was resolving conflicts after the addition of the architecture flag.
Perhaps this could be merged soonish?

Issue #470, opened 2019-07-26
Exception with createTraining using DeLFT


de-code:
I am getting the following error when trying to create training data with DeLFT enabled:

```
Jul 26, 2019 11:31:15 PM org.grobid.core.engines.Engine batchCreateTraining
SEVERE: An error occured while processing the following pdf: /home/deuser/_git_/elife/pdf-xml/data/other/grobid-train-mini-sample/pdf/elife-32671-v2.pdf: org.grobid.core.exceptions.GrobidException: [GENERAL] An exception occurred while running Grobid training data generation for full text.
org.grobid.core.exceptions.GrobidException: [GENERAL] An exception occurred while running Grobid training data generation for full text.
	at org.grobid.core.engines.FullTextParser.createTraining(FullTextParser.java:1348)
	at org.grobid.core.engines.Engine.createTraining(Engine.java:449)
	at org.grobid.core.engines.Engine.batchCreateTraining(Engine.java:529)
	at org.grobid.core.engines.ProcessEngine.createTraining(ProcessEngine.java:379)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.grobid.core.utilities.Utilities.launchMethod(Utilities.java:409)
	at org.grobid.core.main.batch.GrobidMain.main(GrobidMain.java:184)
Caused by: org.grobid.core.exceptions.GrobidException: [GENERAL] An exception occured while running Grobid.
	at org.grobid.core.engines.CitationParser.trainingExtraction(CitationParser.java:743)
	at org.grobid.core.engines.FullTextParser.createTraining(FullTextParser.java:992)
	... 9 more
Caused by: java.lang.NullPointerException
	at org.grobid.core.engines.CitationParser.writeField(CitationParser.java:752)
	at org.grobid.core.engines.CitationParser.trainingExtraction(CitationParser.java:574)
	... 10 more
```

I modified `Engine.java` to show the full exception:

```java
                    LOGGER.error("An error occured while processing the following pdf: "
                        + pdfFile.getPath() + ": " + exp, exp);
```

Generating the training data for the same sample PDF with Wapiti works fine.

/cc @kermitt2 @lfoppiano

PR #471, opened 2019-07-29
create training data: log full exception

de-code:
Logs the full exception rather than just the message.

This helped to narrow down #470


de-code:
> could you please remove the `+ exp`? We don't need it.
> Something like this:
> `LOGGER.error("An error occured while processing the following pdf: "+ pdfFile.getPath(), exp);`
> 
> would be enough

I removed the `exp` from the message itself as requested.

lfoppiano:
Thanks, sorry I could have do it when merging but I though it was quicker for you

de-code:
> Thanks, sorry I could have do it when merging but I though it was quicker for you

I wasn't complaining just notifying ;-) I am the right person to make the change.

lfoppiano:
No worries. I haven't though you did... I just really felt lazy to do it myself 😸

Issue #472, opened 2019-07-29
Exception processing PDF with bad character encoding (0.5.4+)


de-code:
When processing a PDF that has bad character encoding, GROBID now responds with error 500 because it can't parse the generated LXML:

```
ERROR [2019-07-29 15:19:19,034] org.grobid.service.process.GrobidRestProcessFiles: An unexpected exception occurs. 
! org.apache.xerces.impl.io.MalformedByteSequenceException: Invalid byte 2 of 3-byte UTF-8 sequence.
! at org.apache.xerces.impl.io.UTF8Reader.invalidByte(Unknown Source)
! at org.apache.xerces.impl.io.UTF8Reader.read(Unknown Source)
! at org.apache.xerces.impl.XMLEntityScanner.load(Unknown Source)
! at org.apache.xerces.impl.XMLEntityScanner.scanLiteral(Unknown Source)
! at org.apache.xerces.impl.XMLScanner.scanAttributeValue(Unknown Source)
! at org.apache.xerces.impl.XMLDocumentFragmentScannerImpl.scanAttribute(Unknown Source)
! at org.apache.xerces.impl.XMLDocumentFragmentScannerImpl.scanStartElement(Unknown Source)
! ... 80 common frames omitted
! Causing: org.xml.sax.SAXParseException: Invalid byte 2 of 3-byte UTF-8 sequence.
! at org.apache.xerces.util.ErrorHandlerWrapper.createSAXParseException(Unknown Source)
! at org.apache.xerces.util.ErrorHandlerWrapper.fatalError(Unknown Source)
! at org.apache.xerces.impl.XMLErrorReporter.reportError(Unknown Source)
! at org.apache.xerces.impl.XMLErrorReporter.reportError(Unknown Source)
! at org.apache.xerces.impl.XMLDocumentFragmentScannerImpl$FragmentContentDispatcher.dispatch(Unknown Source)
! at org.apache.xerces.impl.XMLDocumentFragmentScannerImpl.scanDocument(Unknown Source)
! at org.apache.xerces.parsers.XML11Configuration.parse(Unknown Source)
! at org.apache.xerces.parsers.XML11Configuration.parse(Unknown Source)
! at org.apache.xerces.parsers.XMLParser.parse(Unknown Source)
! at org.apache.xerces.parsers.AbstractSAXParser.parse(Unknown Source)
! at org.apache.xerces.jaxp.SAXParserImpl$JAXPSAXParser.parse(Unknown Source)
[...]

PR #475, opened 2019-08-05
ignore invalid utf-8 sequences by default

de-code:
This is a workaround for #472 

By default it will ignore invalid UTF-8 sequences as it did in GROBID 0.5.3 and before.
The validation can be turned on via the properties file (`grobid.3rdparty.pdf2xml.validation.enabled`).

/cc @lfoppiano

### DIFF
--- a/grobid-core/src/main/java/org/grobid/core/utilities/GrobidProperties.java
+++ b/grobid-core/src/main/java/org/grobid/core/utilities/GrobidProperties.java
@@ -580,6 +580,12 @@ public class GrobidProperties {
         return Integer.parseInt(getPropertyValue(GrobidPropertyKeys.PROP_3RD_PARTY_PDFTOXML_TIMEOUT_SEC, "60"), 10) * 1000;
     }
 
+    public static boolean isPdfToXMLValidationEnabled() {
+        return Utilities.stringToBoolean(
+            getPropertyValue(GrobidPropertyKeys.PROP_3RD_PARTY_PDFTOXML_VALIDATION_ENABLED
+        ));
+    }
+
     /**
      * Returns the number of threads, given in the grobid-property file.
      *
--- a/grobid-core/src/main/java/org/grobid/core/utilities/GrobidPropertyKeys.java
+++ b/grobid-core/src/main/java/org/grobid/core/utilities/GrobidPropertyKeys.java
@@ -15,6 +15,7 @@ public interface GrobidPropertyKeys {
     String PROP_3RD_PARTY_PDFTOXML = "grobid.3rdparty.pdf2xml.path";
     String PROP_3RD_PARTY_PDFTOXML_MEMORY_LIMIT = "grobid.3rdparty.pdf2xml.memory.limit.mb";
     String PROP_3RD_PARTY_PDFTOXML_TIMEOUT_SEC = "grobid.3rdparty.pdf2xml.memory.timeout.sec";
+    String PROP_3RD_PARTY_PDFTOXML_VALIDATION_ENABLED = "grobid.3rdparty.pdf2xml.validation.enabled";
 
     String PROP_GROBID_CRF_ENGINE = "grobid.crf.engine";
     String PROP_GROBID_DELFT_PATH = "grobid.delft.install";
--- a/grobid-core/src/main/java/org/grobid/core/document/Document.java
+++ b/grobid-core/src/main/java/org/grobid/core/document/Document.java
@@ -35,6 +35,7 @@ import org.grobid.core.sax.*;
 
 import org.grobid.core.utilities.BoundingBoxCalculator;
 import org.grobid.core.utilities.ElementCounter;
+import org.grobid.core.utilities.GrobidProperties;
 import org.grobid.core.utilities.LayoutTokensUtil;
 import org.grobid.core.utilities.Pair;
 import org.grobid.core.utilities.TextUtilities;
@@ -43,13 +44,22 @@ import org.grobid.core.utilities.matching.EntityMatcherException;
 import org.grobid.core.utilities.matching.ReferenceMarkerMatcher;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.xml.sax.InputSource;
+import org.xml.sax.SAXException;
+import org.xml.sax.helpers.DefaultHandler;
 
+import javax.xml.parsers.ParserConfigurationException;
 import javax.xml.parsers.SAXParser;
 import javax.xml.parsers.SAXParserFactory;
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
 import java.io.Serializable;
+import java.nio.charset.Charset;
+import java.nio.charset.CharsetDecoder;
+import java.nio.charset.CodingErrorAction;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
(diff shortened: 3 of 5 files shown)


grobid-core/src/main/java/org/grobid/core/utilities/GrobidPropertyKeys.java
```diff
@@ -15,6 +15,7 @@
     String PROP_3RD_PARTY_PDFTOXML = "grobid.3rdparty.pdf2xml.path";
     String PROP_3RD_PARTY_PDFTOXML_MEMORY_LIMIT = "grobid.3rdparty.pdf2xml.memory.limit.mb";
     String PROP_3RD_PARTY_PDFTOXML_TIMEOUT_SEC = "grobid.3rdparty.pdf2xml.memory.timeout.sec";
+    String PROP_3RD_PARTY_PDFTOXML_VALIDATION_ENABLED = "grobid.3rdparty.pdf2xml.validation.enabled";
```
de-code: I am assuming this also belongs to "third party"? (just trying to be consistent)

grobid-core/src/main/java/org/grobid/core/utilities/GrobidProperties.java
```diff
@@ -569,6 +569,12 @@ public static Integer getPdfToXMLTimeoutMs() {
         return Integer.parseInt(getPropertyValue(GrobidPropertyKeys.PROP_3RD_PARTY_PDFTOXML_TIMEOUT_SEC, "60"), 10) * 1000;
     }
 
+    public static boolean isPdfToXMLValidationEnabled() {
```
de-code: Not quite sure why we spell `Pdf` and `XML`. Java naming conventions is all caps but personally I prefer `Pdf` and `Xml`.

de-code:
> I'm wondering whether we should replace these invalid characters with some placeholder or just remove them.

Yes, I think that would be better. Actually GROBID 0.5.3 was also displaying some odd character.

de-code:
I am also thinking whether there should be an option to save the erroneous XML files? I had to hack the code to do that and inspect the XML. Or is there an option not to delete the temp files?

kermitt2:
So as I mentioned in the issue, it's something wrong with pdfalto. The contract between pdfalto and grobid now is somehow that pdfalto must neutralize all these encoding problems before sending the ALTO file to grobid.

Of course having grobid robust to that is also very nice. I find the property name misleading because in the context of XML, validation has a different meaning which is XML validation (something we really don't want to do and to allow with the ALTO files), here it's just unicode correctness in the XML parsing (well-formedness). 

What about ignoring the unicode problem all the time in the XML parsing? It's not clear what is the interest of keeping it and having an option to allow it or not.

About the temporary ALTO XML files (the `.lxml` extension), they are indeed always removed. For inspecting those files, I was simply re-generating them with the `pdfalto` command line, which is something very simple and avoid adding options?

de-code:
I have revised it to replace with question marks. It didn't actually completely ignore the bytes but was skipping bytes until the utf-8 sequence is valid. It is now displaying question marks for every invalid byte. In the example above it ends up with `??5` in the output.

grobid-core/src/main/java/org/grobid/core/utilities/GrobidPropertyKeys.java
kermitt2: no it's the XML parsing part inside grobid... not third party, but I would suggest to be robust wrt encoding all the time, so not introduce this additional property.

grobid-core/src/main/java/org/grobid/core/utilities/GrobidProperties.java
kermitt2: Not quite sure neither :) 
Maybe a compromise between the different convention :D

de-code:
> So as I mentioned in the issue, it's something wrong with pdfalto. The contract between pdfalto and grobid now is somehow that pdfalto must neutralize all these encoding problems before sending the ALTO file to grobid.

I would agree that the XML generated by pdfalto should ideally be valid.

> Of course having grobid robust to that is also very nice. I find the property name misleading because in the context of XML, validation has a different meaning which is XML validation (something we really don't want to do and to allow with the ALTO files), here it's just unicode correctness in the XML parsing (well-formedness).

I agree, `validate` is an overloaded term and in the XML world usually associated with schema validation like you said. I couldn't think of a better name, I didn't try very hard to be honest. Any suggestion?

> What about ignoring the unicode problem all the time in the XML parsing? It's not clear what is the interest of keeping it and having an option to allow it or not.

In principal, rejecting badly encoded files seems to be a fair choice. This is really a work-around as it is not clear what to do with the badly encoded bytes. Once `pdfalto` fulfils the contract it might, some people might prefer to turn it on. But I wouldn't entirely opposed to removing the configuration.

> About the temporary ALTO XML files (the `.lxml` extension), they are indeed always removed. For inspecting those files, I was simply re-generating them with the `pdfalto` command line, which is something very simple and avoid adding options?

Okay, fair enough. It is just that it tells me there is an error with that particular XML file but by the time I could look at it, it's gone. Running it manually I would need to make sure that I call it with the same parameters. But it's not a big issue. Let's leave it out.

kermitt2:
> I have revised it to replace with question marks. It didn't actually completely ignore the bytes but was skipping bytes until the utf-8 sequence is valid. It is now displaying question marks for every invalid byte. In the example above it ends up with ??5 in the output.

ok it has to be consistent I think with the way place holders from pdfalto are managed in GROBID, otherwise it's going to be messy. 

pdfalto generates place holders with certain unicode character range (in theory it should do the same in this case too, but for some reason here there's a problem). For me, the way those place holders are used will depend on the application. For instance the place holder could be considered in the further processing because they are structurally useful (for instance in dictionaries, we have special characters which are separators but failing encoding resolution), or we could replace them all with a special character, or we could ignore them all. In grobid-core, we could ignore them I think. 

I don't like putting hard coded question marks. How can we differentiate a normal question mark from an encoding issue then? 

I think the best would be to always ignore encoding issue when parsing XML as you introduced here (we don't want to reject badly encoded ALTO files while we can process them), and just log the encoding issue so that we keep track of the error and address the problem in `pdfalto`.    

I would simply ignore those characters in grobid-core - for instance calling a unique method that indicates what to do for all place holder and "CodingErrorAction"?

de-code:
> > I have revised it to replace with question marks. It didn't actually completely ignore the bytes but was skipping bytes until the utf-8 sequence is valid. It is now displaying question marks for every invalid byte. In the example above it ends up with ??5 in the output.
> 
> ok it has to be consistent I think with the way place holders from pdfalto are managed in GROBID, otherwise it's going to be messy.
> 
> pdfalto generates place holders with certain unicode character range (in theory it should do the same in this case too, but for some reason here there's a problem). For me, the way those place holders are used will depend on the application. For instance the place holder could be considered in the further processing because they are structurally useful (for instance in dictionaries, we have special characters which are separators but failing encoding resolution), or we could replace them all with a special character, or we could ignore them all. In grobid-core, we could ignore them I think.

Using a special character that is unlikely to be used otherwise seems to make sense to me. In some cases the original character could have been important. In the example document it is `μ` (micro) as a unit. What the impact of the misinterpretation of the invalid byte / character is, may not even be clear to the application but only the user. Using that special character could allow it to be visible to the user unless the application really knows what to do with it (and it could replace it).

> I don't like putting hard coded question marks. How can we differentiate a normal question mark from an encoding issue then?

I agree, a question mark is not the best choice. Maybe use a designated unicode character. I don't know what pdfalto is using?

> I think the best would be to always ignore encoding issue when parsing XML as you introduced here (we don't want to reject badly encoded ALTO files while we can process them), and just log the encoding issue so that we keep track of the error and address the problem in `pdfalto`.
> 
> I would simply ignore those characters in grobid-core - for instance calling a unique method that indicates what to do for all place holder and "CodingErrorAction"?

`CodingErrorAction` also has [REPORT](https://docs.oracle.com/javase/8/docs/api/java/nio/charset/CodingErrorAction.html#REPORT) option:

> Action indicating that a coding error is to be reported, either by returning a CoderResult object or by throwing a CharacterCodingException, whichever is appropriate for the method implementing the coding process.

I am not quite sure how to control that it should log and ignore. Do you?

de-code:
I didn't easily find which placholder character used by pdfalto.

Could we maybe agree to for now change it to?:
- use the placeholder character used by pdfalto (to be defined)
- remove the configuration option an always replace invalid utf-8 byte sequences with it

Ideally this shouldn't happen anyway and I raised https://github.com/kermitt2/pdfalto/issues/68

kermitt2:
I've merged manually this PR without the property (commit 2bd9f0006e6f46ffa17fd43b2295a6e036391ba1) - the invalid URF-8 characters are always ignored to avoid failing on the whole document because of a couple crappy character codes :) 
Thanks a lot @de-code !

Issue #476, opened 2019-08-07
Duplicated abstract


lfoppiano:
I have a paper: 

[JPS081033701-CC.pdf](https://github.com/kermitt2/grobid/files/3475347/JPS081033701-CC.pdf)

and when processing the fulltext the abstract and part of the body are duplicated. In particular it looks like as the abstract, extracted 

```
    public String getHeaderFeatured(boolean getHeader,
                                    boolean withRotation) {
        if (getHeader) {
            String theHeader = getHeader();
            if ((theHeader == null) || (theHeader.trim().length() <= 1)) {
                theHeader = getHeaderLastHope();
            }
//System.out.println(theHeader);
        }
```

is basically the whole first page, however this is also extracted furhter after from the fulltext parser (`featSeg = getBodyTextFeatured(doc, documentBodyParts);` returns part of the overlapping abstract and first page). 

As a result I have the abstract and part of the first page duplicated. 

I'm wondering if there is a way to recognise when such event happens and just move everything in the body. 

In this paper there is no `abstract` mention but the layout should help recognizing the two sections

kermitt2:
Hello ! 

This behavior has been discussed in #407 #373 #281 

This method `getHeaderFeatured()` is supposed to be removed at some point and I am not working on it anymore since many years. Transition is not very efficient as you can see with #136 which keep track of moving the the new header model (started in 2016) because it requires some time and focus. 

So it's not really an error, it's more a "compromise"/fallback due to the current weakness of the segmentation model approach to locate the header: 
- the heuristics approach (with old `getHeaderFeatured()`) still performs a bit better, so I left it as first way to identify the header zone
- using the segmentation model approach as fallback when the heuristics clearly fails also improves a bit the general accuracy, so it is used as second step
- given that we might have 2 different header zones identified, I am taking the largest ones to avoid loosing piece of the document and it might result in duplications of text (because it's better to have two times the same piece of text than one piece totally missing)

As this is basically a hack to cover the fact that we don't have enough training data for the segmentation model, I would really not spend time on hacking the hack, nor work on the old heuristics, but rather improve the segmentation model identification of the header with better  and more training data (better = using the reading order option in pdfalto).   

[...]

PR #479, opened 2019-08-07
optionally disable header heuristics

de-code:
see discussion in #476

/cc @lfoppiano @kermitt2

### DIFF
--- a/grobid-core/src/main/java/org/grobid/core/engines/Engine.java
+++ b/grobid-core/src/main/java/org/grobid/core/engines/Engine.java
@@ -372,8 +372,12 @@ public class Engine implements Closeable {
             result = new BiblioItem();
         }
 
-        Pair<String, Document> resultTEI = parsers.getHeaderParser().processing2(inputFile, result, config);
-		//Pair<String, Document> resultTEI = parsers.getHeaderParser().processing(inputFile, result, config);
+        Pair<String, Document> resultTEI;
+        if (GrobidProperties.isHeaderUseHeuristics()) {
+            resultTEI = parsers.getHeaderParser().processing2(inputFile, result, config);
+        } else {
+            resultTEI = parsers.getHeaderParser().processing(new File(inputFile), result, config);
+        }
         Document doc = resultTEI.getRight();
         return resultTEI.getLeft();
     }
--- a/grobid-core/src/main/java/org/grobid/core/engines/FullTextParser.java
+++ b/grobid-core/src/main/java/org/grobid/core/engines/FullTextParser.java
@@ -133,7 +133,9 @@ public class FullTextParser extends AbstractParser {
             // header processing
             BiblioItem resHeader = new BiblioItem();
             Pair<String, LayoutTokenization> featSeg = null;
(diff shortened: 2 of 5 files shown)


grobid-core/src/main/java/org/grobid/core/engines/Engine.java
```diff
@@ -372,8 +372,12 @@ public String processHeader(String inputFile, GrobidAnalysisConfig config, Bibli
...
 
-        Pair<String, Document> resultTEI = parsers.getHeaderParser().processing2(inputFile, result, config);
-		//Pair<String, Document> resultTEI = parsers.getHeaderParser().processing(inputFile, result, config);
+        Pair<String, Document> resultTEI;
+        if (GrobidProperties.isHeaderUseHeuristics()) {
+            resultTEI = parsers.getHeaderParser().processing2(inputFile, result, config);
+        } else {
+            resultTEI = parsers.getHeaderParser().processing(new File(inputFile), result, config);
+        }
         Document doc = resultTEI.getRight();
```
de-code: what is the purpose of calling `getRight`? We don't seem to use `doc`.

grobid-core/src/main/java/org/grobid/core/engines/Engine.java
```diff
@@ -372,8 +372,12 @@ public String processHeader(String inputFile, GrobidAnalysisConfig config, Bibli
             result = new BiblioItem();
         }
 
-        Pair<String, Document> resultTEI = parsers.getHeaderParser().processing2(inputFile, result, config);
-		//Pair<String, Document> resultTEI = parsers.getHeaderParser().processing(inputFile, result, config);
+        Pair<String, Document> resultTEI;
+        if (GrobidProperties.isHeaderUseHeuristics()) {
+            resultTEI = parsers.getHeaderParser().processing2(inputFile, result, config);
```
de-code: I am scared to rename `processing2` and `processing`..

grobid-core/src/main/java/org/grobid/core/engines/Engine.java
kermitt2: very likely some old stuff not used any more!

kermitt2:
We want to allow by default the heuristics, no? It performs better in average and most of the users will still benefit from it until some efforts are made on the segmentation model.  
Not using the heuristics would be more at this stage for the power users without fear !

de-code:
> We want to allow by default the heuristics, no? It performs better in average and most of the users will still benefit from it until some efforts are made on the segmentation model.
> Not using the heuristics would be more at this stage for the power users without fear !

Okay, I think I misunderstood you. You seemed keen to disable it all together. Leaving it as it is now is the safer option. I will change the default.

de-code:
I was going to add `Engine` and `FullTextParser` tests but I think it would be easier if they were broken up more. Probably not worth it.

kermitt2:
May I merge now ?

de-code:
> May I merge now ?

Yes, please. That would be a new record :D

PR #485, opened 2019-08-19
Duplication of abstract / body part

lfoppiano:
An attempt to fix the #476

### DIFF
--- a/grobid-core/src/main/java/org/grobid/core/engines/FullTextParser.java
+++ b/grobid-core/src/main/java/org/grobid/core/engines/FullTextParser.java
@@ -79,17 +79,17 @@ public class FullTextParser extends AbstractParser {
 //    private String pathXML = null;
 //	private BiblioItem resHeader = null;
 
-	// default bins for relative position
-	private static final int NBBINS_POSITION = 12;
+    // default bins for relative position
+    private static final int NBBINS_POSITION = 12;
 
-	// default bins for inter-block spacing
-	private static final int NBBINS_SPACE = 5;
+    // default bins for inter-block spacing
+    private static final int NBBINS_SPACE = 5;
 
-	// default bins for block character density
-	private static final int NBBINS_DENSITY = 5;
+    // default bins for block character density
+    private static final int NBBINS_DENSITY = 5;
 
-	// projection scale for line length
-	private static final int LINESCALE = 10;
+    // projection scale for line length
+    private static final int LINESCALE = 10;
 
     private EngineParsers parsers;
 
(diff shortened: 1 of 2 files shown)


lfoppiano:
Yes, I created a new branch + PR because I wasn't sure I wasn't breaking everything :-)

kermitt2:
I'll try to redo some tests with PMC 1942 to review what's the impact on benchmarking. As all these fallback/rewrite/heuristics are only motivated by average accuracy improvement, I think it's better to follow the benchmarking to decide which one to choose - which also means that they will always introduce on the other hand some errors in particular cases :/ 

On the other hand, it might be good to consider a better evaluation set than PMC 1942, with more variety of domains and publishers, possibly mixing some available ISTEX XML fulltext with PMC.

lfoppiano:
@kermitt2 before, please give me another day or two to double check my changes (also in relation to your comments on #476) - I've made changes in the logic and it might not correct anymore.

lfoppiano:
I've implemented a more conservative approach - if I did not overlooked something - that should iterate over the list of tokens (in this case of the abstract) and fetch the document part related to it. 

Once the problem is fixed here, this method can be rewritten to minimise the number of DocumentPieces. 

The issue is still not solved unfortunately, although my abstract has 266 tokens, the result of `getBodyTextFeatured()` is overflowing the abstract, creating the duplications. 

This method is quite complex to follow - the features generation is also here - I feel some rewriting might be necessary here..

kermitt2:
As this PR mixes two very different things, 1) possible change about the usage of the embedded metadata in the PDF and 2) duplication of body part in the abstract - I open two different PR for each point, given also that the tests are different. 

I also only focus on the changes related to the issues, because right now it's very hard (not far from impossible) to follow and review the changes because everything in the file is reformatted and it's very time consuming to check all these changes (around 1000 edit places, if you had doubts about turning off this auto-reformatting, this PR is pretty convincing I think :D ). 

It's better to open other PR with just the reformatting I think, not mixing with issues to be reviewed.

In the part which reprocess the abstract, if I am understanding well, one DocumentPiece per layout token is created, which has several problems. The idea of the DocumentPiece is to identify an area of the document without using additional lists of LayoutToken which are taking quite a lot of memory and extra processing (given that at the segmentation level, we have possibly lists with thousand elements).

The main problem is that the elements not part of the abstract, but surrounded with the pieces of the abstract (the problematic non-continuous abstract) will still be added, so the same problem as before will occur here I think.  

In the dedicated PR for this part, I put a fix for this part with minimal number of DocumentPiece corresponding to continuous layout tokens, which works fine on the example PDF.

lfoppiano:
> As this PR mixes two very different things, 1) possible change about the usage of the embedded metadata in the PDF and 2) duplication of body part in the abstract - I open two different PR for each point, given also that the tests are different.

OK no problem - I had some other tasks and I wanted to be able to see results quickly. 

> I also only focus on the changes related to the issues, because right now it's very hard (not far from impossible) to follow and review the changes because everything in the file is reformatted and it's very time consuming to check all these changes (around 1000 edit places, if you had doubts about turning off this auto-reformatting, this PR is pretty convincing I think :D ).

sorry again. The reformatting is not automatic as a setting in the IDE, is just a uncontrolled reflex in my hands, I do invoke it without even noticing anymore. Normally I notice that when I review the changes and commit only the relevant parts. This time I forgot. 

> It's better to open other PR with just the reformatting I think, not mixing with issues to be reviewed.

> In the part which reprocess the abstract, if I am understanding well, one DocumentPiece per layout token is created, which has several problems. The idea of the DocumentPiece is to identify an area of the document without using additional lists of LayoutToken which are taking quite a lot of memory and extra processing (given that at the segmentation level, we have possibly lists with thousand elements).
> 
> The main problem is that the elements not part of the abstract, but surrounded with the pieces of the abstract (the problematic non-continuous abstract) will still be added, so the same problem as before will occur here I think.
> 
> In the dedicated PR for this part, I put a fix for this part with minimal number of DocumentPiece corresponding to continuous layout tokens, which works fine on the example PDF.

Yes that was just a bad attempt. I admit I did not really understand how the DocumentPiece work... 

We can close this PR and work on the other two you created.

PR #486, opened 2019-08-20
Avoid duplicated body part in the abstract

kermitt2:
This is a fix for one of the issue raised in #476.

The problem is the selection of the abstract to be given to the fulltext model for further structuring the abstract (paragraphs, citations, ..). The current approach supposes incorrectly that the abstract was a continuous segment of the document, while it can consist off several non continuous parts.  

In the new version, we create several DocumentPiece according to the continuous segments.

1) Processing of this PDF is now working (without heuristics):

[EPL0580589-CC.pdf](https://github.com/kermitt2/grobid/files/3520608/EPL0580589-CC.pdf)

2) Test on the 1942 PMC dataset improves abstract identification quite significantly for the fuzzy matches, which is nice:

Before:

```
==== Levenshtein Matching ===== (Minimum Levenshtein distance at 0.8)

abstract             94.11        78.41        72.21        75.18 
```

After:

```
==== Levenshtein Matching ===== (Minimum Levenshtein distance at 0.8)

abstract             95.84        87.93        80.43        84.01  
```

### DIFF
--- a/grobid-core/src/main/java/org/grobid/core/engines/FullTextParser.java
+++ b/grobid-core/src/main/java/org/grobid/core/engines/FullTextParser.java
@@ -162,18 +162,52 @@ public class FullTextParser extends AbstractParser {
                     }
                 }
             }
+
             // structure the abstract using the fulltext model
             if ( (resHeader.getAbstract() != null) && (resHeader.getAbstract().length() > 0) ) {
                 List<LayoutToken> abstractTokens = resHeader.getLayoutTokens(TaggingLabels.HEADER_ABSTRACT);
                 if ( (abstractTokens != null) && (abstractTokens.size()>0) ) {
                     SortedSet<DocumentPiece> documentParts = new TreeSet<DocumentPiece>();
-                    int endInd = abstractTokens.size()-1;
-                    int posStartAbstract = getDocIndexToken(doc, abstractTokens.get(0));
-                    int posEndAbstract = getDocIndexToken(doc, abstractTokens.get(endInd));
-                    DocumentPointer dp1 = new DocumentPointer(doc, abstractTokens.get(0).getBlockPtr(), posStartAbstract);
-                    DocumentPointer dp2 = new DocumentPointer(doc, abstractTokens.get(endInd).getBlockPtr(), posEndAbstract);
-                    DocumentPiece piece = new DocumentPiece(dp1, dp2);
-                    documentParts.add(piece);
+                    // identify continuous sequence of layout tokens in the abstract
+                    int posStartAbstractPiece = -1;
+                    int currentOffsetAbstract = -1;
+                    int startBlockPtr = -1;
+                    LayoutToken previousAbstractToken = null;
+                    for(LayoutToken abstractToken : abstractTokens) {
+                        if (currentOffsetAbstract == -1) {
+                            currentOffsetAbstract = abstractToken.getOffset();
+                            posStartAbstractPiece = getDocIndexToken(doc, abstractToken);
+                            startBlockPtr = abstractToken.getBlockPtr();
+                        } else {
+                            if (abstractToken.getOffset() != currentOffsetAbstract + previousAbstractToken.getText().length()) {
+                                // new DocumentPiece to be added 
+                                DocumentPointer dp1 = new DocumentPointer(doc, startBlockPtr, posStartAbstractPiece);
+                                DocumentPointer dp2 = new DocumentPointer(doc, 
+                                    previousAbstractToken.getBlockPtr(), 
+                                    getDocIndexToken(doc, previousAbstractToken));
+                                DocumentPiece piece = new DocumentPiece(dp1, dp2);
+                                documentParts.add(piece);
+
+                                // set index for the next DocumentPiece
+                                currentOffsetAbstract = abstractToken.getOffset();
+                                posStartAbstractPiece = getDocIndexToken(doc, abstractToken);
+                                startBlockPtr = abstractToken.getBlockPtr();
+                            } 
+                        }
+                        currentOffsetAbstract = abstractToken.getOffset();
+                        previousAbstractToken = abstractToken;
+                    }
+                    // we still need to add the last document piece
+                    // conditional below should always be true because abstract is not null if we reach this part, but paranoia is good when programming 
+                    if (posStartAbstractPiece != -1) {
+                        DocumentPointer dp1 = new DocumentPointer(doc, startBlockPtr, posStartAbstractPiece);
+                        DocumentPointer dp2 = new DocumentPointer(doc, 
+                            previousAbstractToken.getBlockPtr(), 
+                            getDocIndexToken(doc, previousAbstractToken));
+                        DocumentPiece piece = new DocumentPiece(dp1, dp2);
+                        documentParts.add(piece);
+                    }
+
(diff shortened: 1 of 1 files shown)


lfoppiano:
Indeed this change it was done properly. 👍 

I have a couple of minor suggestions. 
I've noticed the same (old version) code was used in the method `processShort()`. I suggest to move the code in a separate method and reuse it there to be on a safe side: 

   ```java
        SortedSet<DocumentPiece> documentParts = new TreeSet<>();

        for (List<LayoutToken> chunk : tokenChunks) {
            documentParts.addAll(collectPiecesFromLayoutTokens(doc, chunk));
        }
   ```

I have already implemented it while I was checking it, I can push them on this branch or send the patch - no reformatting I promise.

lfoppiano:
I notice that disabling the further abstract reformatting returns slightly different text (with the reformatting, there are many more spaces after punctuation). 

In the next example [EPJ0290369-CC.pdf](https://github.com/kermitt2/grobid/files/3523158/EPJ0290369-CC.pdf) the first line is without reformatting:

```
<p>A large positive magnetoresistivity (up to tens of percents) is observed in both underdoped and overdoped superconducting La 2-x Sr x CuO 4 epitaxial thin films at temperatures far above the superconducting critical temperature T c. For the underdoped samples, this magnetoresistance far above T c cannot be described by the Kohler rule and we believe it is to be attributed to the influence of superconducting fluctuations. In the underdoped regime, the large magnetoresistance is only present when at low temperatures superconductivity occurs. T he strong magnetoresistivity, which persists even at temperatures far above T c , can be related to the pairs forming eventually the superconducting state below T c. Our observations support the idea of a close relation between the pseudogap and the superconducting gap and provide new indications for the presence of pairs above T c .</p>
<p>A large positive magnetoresistivity (up to tens of percents) is observed in both underdoped and overdoped superconducting La 2-x Sr x CuO 4 epitaxial thin films at temperatures far above the superconducting critical temperature T c . For the underdoped samples, this magnetoresistance far above T c cannot be described by the Kohler rule and we believe it is to be attributed to the influence of superconducting fluctuations. In the underdoped regime, the large magnetoresistance is only present when at low temperatures superconductivity occurs. T he strong magnetoresistivity, which persists even at temperatures far above T c , can be related to the pairs forming eventually the superconducting state below T c . Our observations support the idea of a close relation between the pseudogap and the superconducting gap and provide new indications for the presence of pairs above T c .</p>
```

I'm not sure where these differences are done...

kermitt2:
> I've noticed the same (old version) code was used in the method processShort().

Ah yes I completely forgot that I've already written exactly the same piece of program before :D 

The idea of `processShort()` was indeed to apply the full text model to any sequence of LayoutToken in a document, it was used for figure and table captions - so we should reuse that for the abstract. I don't remember why I've not reused it when adding the processing of abstract (probably forgot). 

The change would be then simply something like that: 

```
if ( (abstractTokens != null) && (abstractTokens.size()>0) ) {
      Pair<String, List<LayoutToken>> abstractProcessed = processShort(abstractTokens, doc);
      resHeader.setLabeledAbstract(abstractProcessed.getLeft());
      resHeader.setLayoutTokensForLabel(abstractText.getRight(), TaggingLabels.HEADER_ABSTRACT);
}
```

That's it, I think.

lfoppiano:
Ok I see, although in `processShort` you create chunks of layout token and then create document pieces based on that. Yes is the same result... 

Tomorrow I will test it and push the final version now that is more clear 😃

kermitt2:
Yes the new version is actually better, it avoids these extra chunks of layout tokens, so better to move it to `processShort()` I think.

lfoppiano:
OK!

kermitt2:
some loss with the benchmarking, the last changes need to be re-worked

grobid-core/src/main/java/org/grobid/core/engines/FullTextParser.java
```diff
@@ -162,18 +162,66 @@ public Document processing(DocumentSource documentSource,
                     }
                 }
             }
+
             // structure the abstract using the fulltext model
             if ( (resHeader.getAbstract() != null) && (resHeader.getAbstract().length() > 0) ) {
+                List<LayoutToken> abstractTokens = resHeader.getLayoutTokens(TaggingLabels.HEADER_ABSTRACT);
+                if ( (abstractTokens != null) && (abstractTokens.size()>0) ) {
+                    if ( (abstractTokens != null) && (abstractTokens.size()>0) ) {
```
lfoppiano: there is a duplicated `if`, I would also rewrite this part directly with CollectionUtils and StringUtils to improve readability: 
```
            // structure the abstract using the fulltext model
            if ( isNotBlank(resHeader.getAbstract())) {
                List<LayoutToken> abstractTokens = resHeader.getLayoutTokens(TaggingLabels.HEADER_ABSTRACT);
                if (CollectionUtils.isNotEmpty(abstractTokens)) {
                    Pair<String, List<LayoutToken>> abstractProcessed = processShort(abstractTokens, doc);
                    if (abstractProcessed != null) {
                        resHeader.setLabeledAbstract(abstractProcessed.getLeft());
                        resHeader.setLayoutTokensForLabel(abstractProcessed.getRight(), TaggingLabels.HEADER_ABSTRACT);
                    }
                }
            }
```

grobid-core/src/main/java/org/grobid/core/engines/FullTextParser.java
```diff
@@ -300,8 +348,72 @@ else if (config.getConsolidateCitations() == 2)
...
+            if (isNotEmpty(trim(processedText))) 
+                res = label(processedText);
+        } else
+            return null;
+
+        return Pair.of(res, tokenizations);
+    }
+
+
     public Pair<String, List<LayoutToken>> processShort(List<LayoutToken> tokens, Document doc) {
```
lfoppiano: I'm not sure, but in this method `processShort` there is `if (currentPos+1 == tokenPos) {`. 

I have a doubt, but looking at the result of 
```GrobidAnalyzer.getInstance().tokenizeWithLayoutToken("This is a sentence")``` 

I get that `GrobidAnalyzer.getInstance().tokenizeWithLayoutToken("This is a sentence").get(1).getOffset()` is `4`, (`0 + lenght(This)`) so I'm not sure the `+1` is need there... two continuous tokens should be just `previousOffset + previousLenght == currentOffset`. 

Does it make sense? 

As a side note, I've discovered it when trying to write a unit test on the code that identifies the DocumentPieces. If you want I can push it (together with having moved the specific code that calculate the `DocumenPieces` on a separate method).

lfoppiano: I have somehow ended up again on this problem, and usually `currentPos + textLenght == nextToken.getOffset()` 🙂

lfoppiano:
I've pushed an updated version where I've done the following: 
1. move the code to collect the document pieces in a separate method called `collectPiecesFromLayoutToken()`
2. refined the methods 
 - processShort() / old version, working first on the layout token and creating document pieces from it  and 
 - processShortNew() / new version using the layout token list as it is and apply collectPiecesFromLayoutToken() to extract subsequent document pieces only from subsequent parts
3. added one unit test - not enough time unfortunately 
4. I've corrected processShort index as commented in a previous review

Hope this can be accepted soon

Issue #484, opened 2019-08-19
Title/Authors are overwritten using the PDF metadata version


lfoppiano:
I'm investigating some issues with duplicated text when disabling the heuristic approach for header extraction and I've noticed the following, in `FullTextParser.java:148`:  

``` 
                    if (metadata.getTitle() != null)
                        resHeader.setTitle(metadata.getTitle());
```

In principle it make sense, but in this case I get `Absence of pseudogap in heavily overdoped Bi 2 Sr 2 CaCu 2 O 8+δ from tunneling spectroscopy of break junctions` overwritten by `7074.dvi`

Is not a trivial task how to merge the two information, I wonder whether there is a better approach. The same happens to authors, btw.

Original PDF: 
[EPL0580589-CC.pdf](https://github.com/kermitt2/grobid/files/3514576/EPL0580589-CC.pdf)

lfoppiano:
Maybe the idea was to override only if `resHeader.getTitle()` was null or empty?

kermitt2:
Yes that's the idea, title and authors are set via the provided pdf embedded metadata in case the other approaches failed. In average with PMC 1942 dataset, it helps a bit, it's why it has been introduced. 

Maybe with better reading order and a reviewed header model, this won't be relevant anymore in the future. 

About the criteria to apply this fall back, we have:

```
if ((resHeader.getTitle() == null) || (resHeader.getTitle().trim().length() == 0) ||
                 (resHeader.getAuthors() == null) || (resHeader.getFullAuthors() == null) ||
                 (resHeader.getFullAuthors().size() == 0) ) 
```

So when there is no author extracted, we use this as a sign of likely failure of the title too, and the embedded metadata are used for the title. This might not be a safe approach in general, through it helps with the evaluation PMC subset (PDF metadata in PMC might be better quality than in average).

lfoppiano:
Yes, the problem is that just after that line, `ResHeader` is updated: 

```
resHeader = new BiblioItem();
                parsers.getHeaderParser().processingHeaderSection(config.getConsolidateHeader(), doc, resHeader);
```

and it could contains the correctly extracted title. 

My idea was to check again if `resHeader.getTitle()` is null before overwriting with the metadata, something like: 

```
if (metadata.getTitle() != null && resHeader.getTitle() == null)
                        resHeader.setTitle(metadata.getTitle());
```

[...]

PR #487, opened 2019-08-20
Possible change for using embedded PDF metadata

kermitt2:
Following issue #484 and initial PR #485, this PR addresses the usage of embedded PDF metadata as fall back to set the title and authors when the other extraction approaches failed (with and without heuristics).

The embedded PDF metadata are provided by pdfalto, there is no guarantee that they correspond to the actual article title and authors, but it's sometime the case.

The best approach, based on the average accuracy improvement, should be evaluated with the PMC 1942 dataset or a better dataset.

Note that we could imagine some sort of post-validation to better control the usage of these PDF metadata.

### DIFF
--- a/grobid-core/src/main/java/org/grobid/core/document/TEIFormatter.java
+++ b/grobid-core/src/main/java/org/grobid/core/document/TEIFormatter.java
@@ -988,7 +988,7 @@ public class TEIFormatter {
                                     StringBuilder tei,
                                     Document doc,
                                     GrobidAnalysisConfig config) throws Exception {
-        List<String> allNotes = new ArrayList<String>();
+        List<String> allNotes = new ArrayList<>();
         for (DocumentPiece docPiece : documentNoteParts) {
             
             List<LayoutToken> noteTokens = doc.getDocumentPieceTokenization(docPiece);
--- a/grobid-core/src/main/java/org/grobid/core/engines/FullTextParser.java
+++ b/grobid-core/src/main/java/org/grobid/core/engines/FullTextParser.java
@@ -145,7 +145,7 @@ public class FullTextParser extends AbstractParser {
                 // above, use the segmentation model result
                 if (doc.getMetadata() != null) {
                     Metadata metadata = doc.getMetadata();
-                    if (metadata.getTitle() != null)
+                    if (metadata.getTitle() != null && resHeader.getTitle() == null)
                         resHeader.setTitle(metadata.getTitle());
                     if (metadata.getAuthor() != null) {
                         resHeader.setAuthors(metadata.getAuthor());
--- a/grobid-core/src/main/java/org/grobid/core/engines/FullTextParser.java
+++ b/grobid-core/src/main/java/org/grobid/core/engines/FullTextParser.java
@@ -2,6 +2,7 @@ package org.grobid.core.engines;
 
 import com.google.common.collect.Iterables;
 
+import org.apache.commons.collections4.CollectionUtils;
 import org.apache.commons.lang3.tuple.Pair;
 import org.apache.commons.io.FileUtils;
 
@@ -52,6 +53,8 @@ import java.io.IOException;
 import java.io.OutputStreamWriter;
 import java.io.Writer;
 
+import java.nio.charset.StandardCharsets;
+import java.text.Normalizer;
 import java.util.Arrays;
 import java.util.ArrayList;
 import java.util.List;
@@ -61,6 +64,7 @@ import java.util.SortedSet;
 import java.util.StringTokenizer;
 import java.util.TreeSet;
 import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import static org.apache.commons.lang3.StringUtils.*;
 
@@ -92,6 +96,8 @@ public class FullTextParser extends AbstractParser {
 
     private EngineParsers parsers;
 
+    private Pattern pattern = Pattern.compile("[^a-zA-Z0-9]+");
+
     /**
      * TODO some documentation...
      */
@@ -103,8 +109,8 @@ public class FullTextParser extends AbstractParser {
 
 	public Document processing(File inputPdf,
 							   GrobidAnalysisConfig config) throws Exception {
-		DocumentSource documentSource = 
-			DocumentSource.fromPdf(inputPdf, config.getStartPage(), config.getEndPage(), 
+		DocumentSource documentSource =
+			DocumentSource.fromPdf(inputPdf, config.getStartPage(), config.getEndPage(),
 				config.getPdfAssetPath() != null, true, false);
 		return processing(documentSource, config);
 	}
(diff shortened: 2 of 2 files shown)


kermitt2:
I've done some tests with the PMC 1942 dataset. Using embedded metadata as a fall back overall is helping a bit, but actually none of the variants looks really significant: 

* without embedded PDF metadata

```
= Ratcliff/Obershelp Matching = (Minimum Ratcliff/Obershelp similarity at 0.95)

authors              98.55        93.73        93.2         93.46  
title                98.92        95.84        94.9         95.37  
```

* with current master (`if (metadata.getTitle() != null)`)

```
= Ratcliff/Obershelp Matching = (Minimum Ratcliff/Obershelp similarity at 0.95)

authors              98.55        93.59        93.2         93.39  
title                98.99        95.61        95.27        95.44  
```

* with suggested change (`if (metadata.getTitle() != null && resHeader.getTitle() == null)`):

```
= Ratcliff/Obershelp Matching = (Minimum Ratcliff/Obershelp similarity at 0.95)

authors              98.54        93.53        93.15        93.34  
title                99           95.51        95.32        95.41 
```

* with conditional on authors (`if (metadata.getAuthor() != null && (resHeader.getAuthors() == null && resHeader.getFullAuthors() == null)) {`):

```
= Ratcliff/Obershelp Matching = (Minimum Ratcliff/Obershelp similarity at 0.95)

authors              98.58        93.69        93.3         93.5   
title                98.99        95.61        95.27        95.44  
```

So almost the same, the last variant being slightly better than the master.

However, it might depend a lot on the dataset and something more heterogeneous would be an improvement.

lfoppiano:
I have though about it and to start with we could implement/improved these two minimal checkes: 

1. check whether the metadata are notBlank instead of not null:  
[JPS0710709-CC.pdf](https://github.com/kermitt2/grobid/files/3523108/JPS0710709-CC.pdf)

2. check whether the title is not a filename, might be quite complicated to avoid false positive. We could start with checking whether it ends with .doc, .pdf, .dvi, .tex for example. 

The issue in general is the heuristic was often giving back a title and the metadata were not used.

kermitt2:
Another one:

3. check if the string present in the metadata matches text in the PDF content (e.g. first two pages), with a soft match (very soft!)

I could re-evaluate the same without the heuristics... 

But looking at the current results, I think we should ignore the metadata for the moment, they are almost not helping (it's not significant) and add complexity for no benefit.

de-code:
> But looking at the current results, I think we should ignore the metadata for the moment, they are almost not helping (it's not significant) and add complexity for no benefit.

But in general, it can be rather confusing when the title extracted may not even be in the document itself. Someone might have copied a previous documented with metadata and forgot to update it.

For evaluation purpose it might be good if it was configurable. Although even then I'd rather turn it off (similar to consolidation).

BTW BioRxiv seem to generally add metadata

lfoppiano:
I've implemented point 1 and 2. The point 3 is implemented but commented. 
The first two changes are simple, they are non breaking and they just avoid having really bad results from the metadata. 

In general with the heuristics active, there should not change as the title and authors are always (or almost always) filled in.

de-code:
Would you be open to making the metadata configurable?

de-code:
If point 3 is potentially breaking, would it make sense to make that configurable as well? I could imagine that there are use-cases where the document itself doesn't include the title but has it in the metadata. (Although not our use-case)

lfoppiano:
> If point 3 is potentially breaking, would it make sense to make that configurable as well? I could imagine that there are use-cases where the document itself doesn't include the title but has it in the metadata. (Although not our use-case)

I didn't spend too much time on point 3, I'm quite in a rush and I prefer to leave it out because what's implemented solve most of the issues. Point 3 might also be slow, we need to spend more time on it, which unfortunately I don't have. 

I would just leave it out.. make it configurable to me looks a bit overkill .. 

my two cents...

de-code:
> > If point 3 is potentially breaking, would it make sense to make that configurable as well? I could imagine that there are use-cases where the document itself doesn't include the title but has it in the metadata. (Although not our use-case)
> 
> I didn't spend too much time on point 3, I'm quite in a rush and I prefer to leave it out because what's implemented solve most of the issues. Point 3 might also be slow, we need to spend more time on it, which unfortunately I don't have.
> 
> I would just leave it out.. make it configurable to me looks a bit overkill ..

Okay, fair enough. In that case probably leave it out entirely if it can't be used.

kermitt2:
I've prepared the changes outside this PR (too many parallel changes in this fulltext method) with 1) and 2), taking the embedded metadata stuff outside the fallback header process block and correcting the additional consolidation conditions:

```java
            // header processing
            BiblioItem resHeader = new BiblioItem();
            Pair<String, LayoutTokenization> featSeg = null;
            if (GrobidProperties.isHeaderUseHeuristics()) {
                // heuristics for identifying the header zone, this is the old version of the header block identification, 
                // still used because more robust than the pure machine learning approach (lack of training data)
                parsers.getHeaderParser().processingHeaderBlock(config.getConsolidateHeader(), doc, resHeader);
            }
            
            if (isBlank(resHeader.getTitle()) || isBlank(resHeader.getAuthors()) || CollectionUtils.isEmpty(resHeader.getFullAuthors())) {
                resHeader = new BiblioItem();
                // using the segmentation model to identify the header zones
                parsers.getHeaderParser().processingHeaderSection(config.getConsolidateHeader(), doc, resHeader);
            }

            if (isBlank(resHeader.getTitle()) || isBlank(resHeader.getAuthors()) || CollectionUtils.isEmpty(resHeader.getFullAuthors())) {
                // try to exploit PDF embedded metadata (the so-called XMP) if we are still without title/authors
                // this is risky as those metadata are highly unreliable, but as last chance, why not :)
                Metadata metadata = doc.getMetadata();
                if (metadata != null) { 
                    boolean titleUpdated = false;
                    boolean authorsUpdated = false;

                    if (isNotBlank(metadata.getTitle()) && isBlank(resHeader.getTitle())) {
                        if (!endsWithAny(lowerCase(metadata.getTitle()), ".doc", ".pdf", ".tex", ".dvi", ".docx", ".odf", ".odt", ".txt")) {
                            resHeader.setTitle(metadata.getTitle());
                            titleUpdated = true;
                        }
                    }

                    if (isNotBlank(metadata.getAuthor())
                        && (isBlank(resHeader.getAuthors()) || CollectionUtils.isEmpty(resHeader.getFullAuthors()))) {
                        resHeader.setAuthors(metadata.getAuthor());
                        resHeader.setOriginalAuthors(metadata.getAuthor());
                        authorsUpdated = true;
                        List<Person> localAuthors = parsers.getAuthorParser().processingHeader(metadata.getAuthor());
                        if (localAuthors != null) {
                            for (Person pers : localAuthors) {
                                resHeader.addFullAuthor(pers);
                            }
                        }
                    }

                    // if title and author have been updated with embedded PDF metadata, we try to consolidate 
                    // again as required 
                    if ( titleUpdated || authorsUpdated ) {
                        parsers.getHeaderParser().consolidateHeader(resHeader, config.getConsolidateHeader());
                    }
                }
            }
```

This new sequence improves finally title and authors on PMC 1942 set a bit. The metadata block improves title from 95.78 to 95.93 (f-score) and authors from 95.56 to 95.72 (first author from 95.83 to 96.25). However, I think we would need a more heterogeneous evaluation set for estimating this more generally (I have one with a couple hundred ISTEX publisher articles but it cannot be made open). 

Issues reported #484 disappear as extented.

kermitt2:
push with commit 8086c8cb0f4e3955896a4af28c1c612e1e17ca5e
latest benchmark:
https://github.com/kermitt2/grobid/blob/master/grobid-trainer/doc/PMC_sample_1943.results.grobid-0.5.5-Glutton-13.09.2019

kermitt2:
Ok I re-ran the evaluations, I really wanted to believe, but there is no improvement at all, I commented by error something else :/
So for the moment, it is better to keep the exploitation of the XMP metadata commented and re-evaluate it in the future with more diverse test data.

kermitt2:
I suggest to close #484 and this PR :) @lfoppiano

lfoppiano:
Yes sure! 👍

PR #488, opened 2019-08-20
added bin to .gitignore

de-code:
Visual Studio Code outputs to the `bin` folder. This would exclude those files from git (not sure if there would be any benefit of including it).


kermitt2:
Yes the files under `bin/` are not used any more, so that should not create any issues.

Issue #495, opened 2019-09-03
superscript/subscript, bold/italics not always parsed correctly


lfoppiano:
I've noticed that bold/italics were taken only from FONTSTYLE, however (as explained here: https://github.com/kermitt2/pdfalto/issues/57#issuecomment-527337488) sometimes the information is present in the font name (ending with `-bold` or `_bd` etc...) 

I've added a unit test and try to fix the issue (https://github.com/kermitt2/grobid/tree/pdfalto_parser_fixes)  in the xml parser, however I'm not sure what's expected to be outputed from pdfalto...

kermitt2:
Font style like bold and italic are already derived from the font name by pdfalto via xpdf (this is common to know about these font attributes), so any style information in FONTSTYLE are already covering the font name and looking again FONTFAMILY in GROBID is redundant. 

In xpdf, I think the main piece performing this check is `GlobalParams.cc, line 237 (version 4.00), this method:

```
SysFontInfo *SysFontList::find(GString *name)
```

So we should never see these attributes missing in FONTSTYLE when present in the font name. If it is the case, this should be fixed normally in xpdf.

lfoppiano:
In this example I have this: 

```
<TextStyle ID="font0" FONTFAMILY="kmhhoc+advtimes_bd" FONTSIZE="11.955" FONTTYPE="serif"
                   FONTWIDTH="proportional" FONTCOLOR="#000000" FONTSTYLE=""/>
        <TextStyle ID="font1" FONTFAMILY="kmhhod+advmtmrb" FONTSIZE="8.368" FONTTYPE="sans-serif" FONTWIDTH="fixed"
                   FONTCOLOR="#000000" FONTSTYLE="subscript"/>
        <TextStyle ID="font2" FONTFAMILY="kmhhof+advmtmib" FONTSIZE="8.368" FONTTYPE="sans-serif" FONTWIDTH="fixed"
                   FONTCOLOR="#000000" FONTSTYLE="subscript"/>
        <TextStyle ID="font3" FONTFAMILY="kmhhog+advtimes_rg" FONTSIZE="9.963" FONTTYPE="serif" FONTWIDTH="proportional"
                   FONTCOLOR="#000000" FONTSTYLE=""/>
        <TextStyle ID="font4" FONTFAMILY="kmhhoh+advmtmr" FONTSIZE="6.974" FONTTYPE="sans-serif"
[...]

PR #496, opened 2019-09-04
Pdfalto parser fixes

lfoppiano:
This should solve the issue #495 and add some more unit tests


lfoppiano:
@Aazhar we've currently failed to build pdfalto for windows. Could you please try to build it and let me know what is the procedure and which libraries should be used? There seems to be a problem with `aconf.h` when using windows + cygwin. (cc @flydutch)

Issue #180, opened 2017-05-03
Hyphen at line break removed


de-code:
In the first pubmed manuscript, a number of times 'α2-integrin' is at a line break, e.g.:
"was mediated through the inhibition of expression of α2-
integrin (1,2). Integrins are receptors that mediate attachment"

In the output it becomes:
"was mediated through the inhibition of expression of α 2integrin..."
(The space is another issue https://github.com/kermitt2/grobid/issues/179)

In some cases it may be desirable to remove the hyphen. Not in this case. Probably never when there is a number?

de-code:
Actually there is another not so clear word hyphenation example:
"Mean PK parameters CL, V, and F were calculated by non-
compartmental analysis. The tumor growth experiments in"

Becomes:
"Mean PK parameters CL, V, and F were calculated by noncompartmental analysis...."

In the nxml file it is annotated as 'non-compartmental'. I believe both versions are valid. So I would probably not treat that as a bug. (But I thought it was worth mentioning anyhow)

kermitt2:
Thanks for the issue!

Dehyphenization is tricky ;). I was aware of these issues, but it's very useful to have a dedicated issue for that and discuss. 

So far what is implemented now is very simplistic - it does not require dictionaries and resources, but it produces the errors you are mentioning. Basically, when there is an hyphen at the end of the line, we always dehyphenize.   

We could introduce additional rules to improve it, like looking if we have numbers in the tokens before concatenating it, and/or to have a language-specific list of prefix (like anti-, non-, post-). We might always have some rare errors/exceptions. Adding rules manually is endless and not really in the spirit of GROBID... 

Another approach could be to use a Machine Learning based text cleaner/normalizer, for instance like `seq2seq`, but then the problem is to have enough training data. The advantage would be maybe to tackle at the same time other text cleaning problems like diacritics combinations, invalid spacing, etc.

de-code:
Intuitively I was going to claim that it isn't common that words are broken across lines. Scanning through the same PDF the evidence shows it is quite common actually. One approach could also be to look at other examples within the same document. I checked a few examples (around 7), the only one I couldn't find another time so far was 'NON-MEM'. If it was for me, then I would like to have the option to include an element for the hyphens at line boundaries. Then I could do some post processing.

Would you be happy to use the PMC dataset as training data?

Otherwise that seems to be training data that should be fairly easily extracted from any PDF with XML / text data (at least in English). Maybe even without PDF.

[...]

PR #498, opened 2019-09-06
Improved dehypenisation

lfoppiano:
I tried to improve the dehypenisation (#180)

Changes: 
 - deprecated all dehypenisation method in TextUtilities (they are "redirected" to the LayoutTokensUtil) 
 - improved recognition using coordinates to check "real" break line 
 - added more Unit tests 

TODO?: use a dictionary to check whether `previousToken + followingToken` are valid words (from dehypeniseHard)

/cc @de-code


kermitt2:
PDF Case of #180 is working fine and the PR is tested against the 1942 PMC PDF without loss, many thanks !

Issue #424, opened 2019-04-17
GROBID 0.5.4 abstract extraction regression


de-code:
There are a few cases where abstracts extracted using GROBID 0.5.4 have regressed compared to 0.5.4.

Most of my examples are author submitted and not public but I found two on BioRxiv:
(19 out of 100 documents have worse results, about 9 have no abstract at all)

- [NECAPs are negative regulators of the AP2 clathrin adaptor complex](https://www.biorxiv.org/content/10.1101/192898v1)
- [Intermittent Ca2+ signals mediated by Orai1 regulate basal T cell motility](https://www.biorxiv.org/content/10.1101/212274v1)

The PDFs contain line numbers but I originally tested it after removing the line numbers.

I am getting the following extraction:

### 0.5.3 

```xml
        <profileDesc>
            <abstract>
                <p>11 Eukaryotic cells internalize transmembrane receptors via clathrin-mediated endocytosis, 12 but it remains unclear how the machinery underpinning this process is regulated. We recently 13 discovered that membrane-associated muniscin proteins such as FCHo and SGIP initiate 14 endocytosis by converting the AP2 clathrin adaptor complex to an open, active conformation 15 that is then phosphorylated (Hollopeter et al., 2014). Here we report that loss of ncap-1, the sole 16 C. elegans gene encoding an adaptiN Ear-binding Coat-Associated Protein (NECAP), bypasses 17 the requirement for FCHO-1. Biochemical analyses reveal AP2 accumulates in an open, 18 phosphorylated state in ncap-1 mutant worms, suggesting NECAPs promote the closed, 19 inactive conformation of AP2. Consistent with this model, NECAPs preferentially bind open and 20 phosphorylated forms of AP2 in vitro and localize with constitutively open AP2 mutants in vivo. 21 NECAPs do not associate with phosphorylation-defective AP2 mutants, implying that 22 phosphorylation precedes NECAP recruitment. We propose NECAPs function late in 23 endocytosis to inactivate AP2.</p>
            </abstract>
        </profileDesc>
```

### 0.5.4

```xml
        <profileDesc>
            <abstract>
                <div
                    xmlns="http://www.tei-c.org/ns/1.0" />
                </abstract>
            </profileDesc>
```

(although I did get an extraction before)

### 0.5.3

```xml
        <profileDesc>
...
            <abstract>
[...]

PR #501, opened 2019-09-10
make labeled abstract configurable

de-code:
related to #424

makes the labeled abstract configurable


de-code:
hmmm looks like I accidentally include the character issues

de-code:
Seems you prefer not to have that configuration and it doesn't seem to be needed anymore. Thank you again.

PR #510, opened 2019-10-16
Parser for Acknowledgments 

tantikristanti:
A branch containing a model for parsing acknowledgment section both for a raw string text and for the Pdf files. Result from the DESIR Code sprint 2.


grobid-trainer/resources/dataset/acknowledgment/evaluation/test.tei.xml
```diff
@@ -0,0 +1,47 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
lfoppiano: If you use the n-fold validation you could move all the data in the training, which will be partitioned before running the test/evaluation sets and then you can use the whole dataset to train, with a bit more data.

grobid-trainer/src/main/java/org/grobid/trainer/AcknowledgmentTrainer.java
```diff
@@ -0,0 +1,158 @@
...
+     * Command line execution.
+     *
+     * @param args Command line arguments.
+     * @throws Exception
+     */
+    public static void main(String[] args) throws Exception {
+        GrobidProperties.getInstance();
+        AcknowledgmentTrainer trainer = new AcknowledgmentTrainer();
+        AbstractTrainer.runTraining(trainer);
+        AbstractTrainer.runEvaluation(trainer);
```
lfoppiano: See comment before. Might not need the evaluation if you use the n-fold validation

lfoppiano (changes requested):
Thank you for this PR. I passed through it and added some comments. Please have a look at them. 

Also why do you need the clearNLP dictionary `data/clearNLP/dictionary-1.3.1.zip`? 

Thank you

grobid-trainer/src/main/java/org/grobid/trainer/AcknowledgmentTrainer.java
tantikristanti: OK, thanks, I will look at it. 
Do you have any documentation to put me on the right path?
Here:
https://grobid.readthedocs.io/en/latest/Training-the-models-of-Grobid/#n-folds-cross-evaluation, will be enough? or 
here 
https://grobid-quantities.readthedocs.io/en/latest/training.html ?

grobid-home/models/date/model.wapiti
```diff
@@ -1,4 +1,4 @@
-#mdl#2#263
```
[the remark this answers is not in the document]
tantikristanti: sorry for this, I retrained several models just as examples but I forgot to return them back before the commit..

grobid-home/models/date/backup/model.wapiti
```diff
@@ -0,0 +1,6921 @@
+#mdl#2#263
```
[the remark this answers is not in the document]
tantikristanti: same reason, I was careless not selecting or deleting the files before commiting

grobid-core/src/main/java/org/grobid/core/engines/FullTextParser.java
```diff
@@ -307,6 +301,10 @@ else if (config.getConsolidateCitations() == 2)
 				//System.out.println(rese);
 			}
 
+			// TODO : acknowledgment like annexes
+            //documentBodyParts = doc.getDocumentPart(SegmentationLabels.ACKNOWLEDGEMENT);
```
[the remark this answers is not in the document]
tantikristanti: not needed, it was only comments during discussions and reminders for me

grobid-core/src/main/java/org/grobid/core/engines/DateParser.java
```diff
@@ -397,6 +397,7 @@ else if (i == ll - 1) {
                     }*/
 
                 String output = writeField(s1, lastTag0, s2, "<day>", "<day>", addSpace, 0);
+                System.out.println(output);
```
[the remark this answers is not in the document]
tantikristanti: yes, I have compared it with the master branch

.travis.yml
```diff
@@ -1,5 +1,6 @@
 language: java
 sudo: true
+dist: trusty
```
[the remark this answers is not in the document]
tantikristanti: yes you are right, I won't need it. Actually I had bugs which made the test in TestFullTextParser failed. I fixed the bugs.

grobid-trainer/src/main/java/org/grobid/trainer/AcknowledgmentTrainer.java
lfoppiano: yes, the first link 

in AbstractTrainer there is a method called `evaluateNFold` or something like that (I'm trying to remember)

grobid-core/src/main/java/org/grobid/core/engines/FullTextParser.java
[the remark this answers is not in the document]
lfoppiano: Ok, just remember to cleanup before sending the PR ;-)

grobid-home/models/acknowledgment/TrainResult.txt
```diff
@@ -1,401 +0,0 @@
-(base) MP16-PRO:grobid tkristan$ ./gradlew train_acknowledgment
```
lfoppiano: this seems the same as the evaluation result - should remove one of the two - ideally this one

lfoppiano (changes requested):

grobid-core/src/main/java/org/grobid/core/data/Acknow.java
```diff
@@ -0,0 +1,33 @@
+package org.grobid.core.data;
+
```
lfoppiano: What is this class for? Could you add a little description?

grobid-core/src/test/java/org/grobid/core/engines/AcknowledgmentParserTest.java
```diff
@@ -31,13 +32,53 @@ public void setUp() throws Exception {
...
+
         final Acknowledgment acknowledgment = output.get(0);
         assertThat(acknowledgment.getFundingAgency(), is("the Deutsche Forschungsgemeinschaft"));
         assertThat(acknowledgment.getProjectName(), is("the SFB 649 \" Economic Risk \""));
 
         //System.out.println(output.get(0).getFundingAgency());
     }
+
+    @Test
+    public void processing_akwnowledgment2_shouldWork() throws Exception {
```
lfoppiano: This test is not testing anything. Please use assertions ;-)

lfoppiano (changes requested):

kermitt2:
Si finally covered with PR #1046, with some re-design, re-annotation and more training data.

Issue #465, opened 2019-07-25
createTraining produces fulltext TEI with unclosed tags


Vitaliy-1:
Tested with the current master branch and latest release.

Problems:
1) unclosed `p` tag
2) redundant `p` tag
3) `figure` element is closed by `</table>` tag. 

@kermitt2, I'll send you PDFs for reproduction.

kermitt2:
Thank you @Vitaliy-1 !

Those non-well-formness problems are quite frequent yes (other tags can be implied). I didn't spend too much time on that so far, because the foreseen purpose of the generated training files was  (until now) to be reviewed manually. 

So I considered correcting well-formness just as one another things that the annotator will correct (something minor actually in comparison to the time needed to fix the labelling).
   
However, this can certainly be fixed quite easily using the Clusteror class now and error cases are very welcomed !

de-code:
The same seem to be happening for the figure training data itself. It randomly contains a `</content>` without the corresponding opening tag.

de-code:
Perhaps related, one of the generated segmentation data had a space in front of the `<?xml` declaration at the top - which is apparently invalid.

lfoppiano:
This should have finally get fixed.

PR #511, opened 2019-10-16
#465 additional figure and paragraph closing tags

Vitaliy-1:
This doesn't fix all the problems but gives an idea what is wrong with the formatting.
The changes from line 1898 are for the situation when markers are at the end of the paragraph, e.g.:
`<p>[Insert <ref type="table">Table 1</ref>   <figure type="table">here]<lb/> [Insert Table 2 here]<lb/></figure>`
Here the `p` tag is not closed although according to the text this data should comprise a whole paragraph. I'm making an additional check, it fixes the problem but I'm not sure if it the right way, e.g. when taking into account lists.
The line 1585 here fixes the problem when a new redundant `p` tag appears inside a paragraph. 
The line 1874 closes `figure type="table"` with an appropriate tag.


kermitt2:
Thank you @Vitaliy-1 and sorry to have been so slow with this PR. 

Due to merging mess, I added manually the changes in 7e69bd1bbd897d5dca4cd49c1585e3e39871740e

PR #515, opened 2019-10-21
Docx support

kermitt2:
docx document support via Apache POI.


kermitt2:
I need to add some tests :/

kermitt2:
I've added examples under ```grobid/grobid-core/src/test/resources/docx```.
Out of 14 different docx article templates, the conversion of 3 are currently failing (`IJSRP-paper-submission-format-double-column.docx`, `IJSRP-paper-submission-format-single-column.docx` and `Journal-Manuscript-Format-MS-Office-2007.docx`) :(

kermitt2:
Given the poor results with Apache POI, we would need to try [docx4j](https://www.docx4java.org/trac/docx4j) as alternative.

Downchuck:
Is there a quick way -- or would it make sense to add the ALTO XML as an input format for the service, as well? In this manner, someone could run that pipeline separately (pdfalto included) and from there just submit the XML.

This is of course not as nice as getting inline converters working, but it does sort of shift the problem.

kermitt2:
Adding web services taking as input ALTO XML instead of PDF is easy (we just shortcut the pdf conversion), but the problem is that there are a certain number of ALTO flavors and so far the only one well tested and supported is the pdfalto output. This is probably quite a lot of effort to test and support comprehensively ALTO variants. 

The best option regarding docx would be certainly a docx to ALTO XML conversion (same ALTO as pdfalto), and just take also ALTO as input, but it looks complicated, which is why I am adding, first, this docx support via Apache POI or docx4j.

Downchuck:
Should I add ALTO XML input as a feature request / pull request?

The embedded PDF ALTO support makes containerization a bit more of a challenge and does have an impact in the way ingestion can be managed- as you could imagine, running the PDF ALTO process on a separate server gives the GROBID server more capacity for running WAIPI and such.

Downchuck:
noting another contributor has posted a pdfalto service PR (WIP): https://github.com/kermitt2/grobid/pull/552

PR #525, opened 2019-12-13
Fixed duplicate word on paragraph 29

kant:


kermitt2:
Thank you @kant for correcting these typos !

PR #533, opened 2020-01-13
update dependencies to remove known security flaws in dependencies of grobid-service

woodfighter:
not extensively tested yet, not a user just a sysad


lfoppiano:
Which kind of tests did you run?

woodfighter:
I used [OWASP Dependency-Check](https://owasp.org/www-project-dependency-check/) with default settings.

lfoppiano:
I tested the interface and everything seems fine.

Only issue, but it's common to the master version - and is not due to this PR, is #548

lfoppiano:
I tested #548 using these changes and the problem is not happening.

PR #538, opened 2020-01-29
Change local unit tests server ports 

lfoppiano:
The port 8075 and 8076 are more likely to be used by other services, when deployed in the same server, causing the tests to fail. I propose to modify them.


kermitt2:
+1 thanks Luca

Issue #410, opened 2019-03-26
Unit test for parser


lfoppiano:
Today I might have found a solution for testing the parsers. I'm writing here hoping to get some feedback. 

We can create a 'special' model call `Models.NONE` (or `Models.DUMMY`), and when passed in the AbstractParser, the Grobid engine just the loading of any of the machinery under the hood (no wapiti/delft, etc)... 
The tests can be initialised by call a second constructor in the parsers (with package visibility)  and passing the dummy identifier for the model which will not require even the grobid-home to be available. 
```
new MyParser(Models.NONE).testMethod1()
```

More in detail `TaggerFactory.getTagger(model)` should return a NoneTagger or DummyTagger which will just return the same input when the method `label()` is called. 

I haven't tried to implement it, in theory seems working ... 

What do you think?

kermitt2:
It looks like a very good simple solution indeed ! 
I don't see right now why it would not work (but there are often surprises D:).  

The only limit would be when we need stuff from the Lexicon (which uses `grobid-home` for init) to generate features, these aspects could then not be tested, which might be a bit limiting in some cases.

lfoppiano:
I've managed to mock the FeatureFactory, for example, using PowerMock, see: 
 - https://github.com/kermitt2/grobid-quantities/blob/master/src/test/java/org/grobid/core/features/FeaturesVectorQuantitiesTest.java
 - https://github.com/kermitt2/grobid-quantities/blob/master/src/test/java/org/grobid/core/features/FeaturesVectorUnitTest.java

lfoppiano:
I think this can be closed, perhaps after I've wrote a page on how to do it in the documentation ;-)

kermitt2:
Oh yes, the developer notes section ! :D

PR #539, opened 2020-01-29
Implementing a dummy-model

lfoppiano:
The advantages: 
 -  possibility to test parser class side methods
 - remove the needs of having the grobid home around (a particular problem for sub-modules)

Implementation: 
 - added a `GrobidModels.DUMMY`
 - added a `DummyTagger` which is returned from TaggerFactory only for `GrobidModels.DUMMY` and tags everything as `<dummy>`

This was mentioned in issue #410


lfoppiano:
Used in the following: https://github.com/lfoppiano/grobid-superconductors/blob/master/src/test/java/org/grobid/core/engines/SuperconductorsParserTest.java 

I will write a few unit tests around it

Issue #514, opened 2019-10-17
Question about identifying table content


jribault:
Hi,

I'm trying to identify patent (with country, number, date.... I have my own model) and it's working pretty well on text but not very well when patent are listed in table. Do you have any advice concerning table recognition ? Should I try to preprocess the text to remove the table ? Should I modify the template ? 

I'm a bit stuck so any advice are welcome :)

Sunnycheey:
My idea is:
First detecting coordinates of tables by GROBID, and then using other tools like tabula to process the desired region.

lfoppiano:
https://grobid.readthedocs.io/en/latest/Frequently-asked-questions/#how-to-extract-table-content-from-pdfs

Issue #340, opened 2018-08-24
Consider integrating Tabula


de-code:
[Tabula](https://tabula.technology/) seem to be actively developed and doesn't perform so badly (haven't done a quantitative analysis yet).

Would you consider integrating it?

kermitt2:
Hi Daniel !

Yes, we (Vincent from CDS Strasbourg @Vi-dot and me) studied the integration of Tabula. 

We started with the idea to integrate [ContentMine module for tables](https://github.com/ContentMine/svg2xml), but we failed to find a reliable way to convert PDF areas (the area identified as table by GROBID) into SVG.

So we then explore tabula-java. Unfortunately looking at the code, we saw that it is completely interlaced with pdfbox for all internal data structures. It mean concretely either: 

- parse again the PDF with PDFBox2, but given that PDFBox is considerably slower than pdf2xml, use much more memories and is less robust, not to say that we will have soon a lot of new low level PDF improvements and robustness with [pdfalto](https://github.com/kermitt2/pdfalto), it would have cancel all these advantages.

- to modify tabula-java for using different internal representations than the one from PDFBox, but the work looks huge. Likely easier then to implement a new method with machine learning within GROBID with our more abstract PDF layout/structure representations.

That's our current state of though about internal table structure parsing...

de-code:
Hi Patrice,

It is great that you two already had a look at it.

Regarding ContentMine, did you have success converting tables with it in general? It did seem a good candidate initially but may not be as actively developed (last time I tried on PKP's coaction dataset the heuristics didn't detect words correctly and there wasn't a table output - but I may have used it incorrectly). In my prototype processing, I was converting _lxml_ (output of your _pdftoxml_) to _SVG_ but I haven't tried to feed it into ContentMine. Just curious which difficulties you encountered? Detecting the correct table area?

For the table integration, I guess it depends how much time you have to keep maintaining it going forward. It might be okay for it to be a bit slower in that case (could be worth to have a configuration to turn it on or off as not everyone will be interested in tables). It might also be good to measure whether GROBID or Tabula is better at detecting tables or the table areas.

Some potential options to consider:

a) Run Tabula in parallel to get the elapsed time down (but take the memory hit, assuming multiple cores are available)
b) Extract the PDF pages with the tables that GROBID detected the tables on
c) Render ALTO (the pages with tables) to PDF which should be easier to parse by Tabula (and should carry over some of the improvements?)

[...]

PR #546, opened 2020-02-25
Parsing of table content

Vitaliy-1:
Algorithms for table parsing. Inspired by: http://ieg.ifs.tuwien.ac.at/projects/pdf2table/yildiz_iicai05.pdf
Particularly, I've implemented the first algorithm from there and the idea in general. I suppose it's called the line-based approach.  It works much better than tabula, as a minimum on my test instance. 
related to: 
https://github.com/kermitt2/grobid/issues/514
https://github.com/kermitt2/grobid/issues/340
Examples: https://vitaliy-1.github.io/tablesExample/

### DIFF
--- a/grobid-core/src/main/java/org/grobid/core/data/Table.java
+++ b/grobid-core/src/main/java/org/grobid/core/data/Table.java
@@ -2,6 +2,10 @@ package org.grobid.core.data;
 
 import org.grobid.core.GrobidModels;
 import org.apache.commons.lang3.StringUtils;
+import org.grobid.core.data.table.Cell;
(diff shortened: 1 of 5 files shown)


kermitt2:
Awesome work @Vitaliy-1 ! I've processed a couple thousand PDF and looked at a few of them. Tables look often very good. The main issue I think, as you noted in your examples, is when the table notes are added in the last line of the table. Possibly we could introduce an additional label in the table model to label those table notes explicitly to get them removed from the table content. What do you think? 

Thanks a lot for the contribution!

PR #552, opened 2020-03-02
expose AltoPdf  as REST API 

jorgeveamurguia:
return xml with text content and all images in base64 format

format XML as : 
<all><content>"+ content +"</content><images>"+ base64imagesxml +"<images><all>

### DIFF
--- a/grobid-core/src/main/java/org/grobid/core/document/DocumentSource.java
+++ b/grobid-core/src/main/java/org/grobid/core/document/DocumentSource.java
@@ -383,7 +383,9 @@ public class DocumentSource {
     public void setXmlFile(File xmlFile) {
         this.xmlFile = xmlFile;
     }
-
+    public void setCleanupXml( boolean cleanupXml ){
+        this.cleanupXml=cleanupXml;
+    }
 }
 
 
--- a/grobid-service/src/main/java/org/grobid/service/AltoRestService.java
+++ b/grobid-service/src/main/java/org/grobid/service/AltoRestService.java
@@ -1,0 +1,45 @@
+package org.grobid.service;
+
+import com.codahale.metrics.annotation.Timed;
+import com.google.inject.Inject;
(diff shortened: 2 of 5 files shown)


lfoppiano:
I'm not sure we want to expose an internal / transportation format at high level such at the API. You write a separate service, that uses grobid-core as a dependency and exposes the information you need or use the information at the programmatic level. 

The ALTO format chained with images in a base64 format seems, IMHO, a bit an ad-hoc solution. 

Could you add some justification of why would you need such service in grobid?

jorgeveamurguia:
It is possible that it is poorly implemented. No new service is necessary. It is better to add a parameter and optionally return image linked to the document. You're right.

I am using Grobid to extract text from PDF documents. But, I would also like to extract images too. So I have implemented a new method and I have exposed it to take image (in BASE64 format).

kermitt2:
Hello @jorgeveamurguia 

You can get the embedded images converted into `.png` by using the batch command. For instance:

```CONSOLE
> java -Xmx4G -jar grobid-core/build/libs/grobid-core-0.6.0-SNAPSHOT-onejar.jar -gH grobid-home -dIn ~/test/in0/ -dOut ~/test/out0 -exe processFullText 
```

Screenshot from 2020-03-09 21-09-50

There is a web service that is doing the same, returning everything in a big zip file, `processFulltextAssetDocument`, still usable but **deprecated**.

But if you are simply interested in raw text and images, you'd better use command lines with https://github.com/kermitt2/pdfalto

jorgeveamurguia:
Hello  @kermitt2 
I want to extract images and texts. I want text to be related to image. 
Position of text and position of image in text is important to me,

lfoppiano:
I close this, as it has been documented [here](https://grobid.readthedocs.io/en/latest/Frequently-asked-questions/#i-would-also-like-to-extract-images-from-pdfs)

PR #554, opened 2020-03-08
Fix LaTeX sytnax of year range

koppor:
Number ranges are typeset in LaTeX using `number--number`. This is implemented here.


kermitt2:
Thanks @koppor !

PR #556, opened 2020-03-09
Create CHANGELOG.md

koppor:
I really like <https://keepachangelog.com/en/1.0.0/>.

Thus, I moved the contents of `Readme.md` to `CHANGELOG.md` and refined it with the contents available at the GitHub tags.

Additionally linted `Readme.md` using <https://github.com/DavidAnson/vscode-markdownlint>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,195 @@
+# Changelog
+
+All notable changes to this project will be documented in this file.
+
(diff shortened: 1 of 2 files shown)


kermitt2:
Thank you @koppor ! Yes this is really nice, very good change. I will update a bit the changelog after merging. 

(I have to say I really like the [ADR](https://adr.github.io/madr/) too, and I will try to make some in the future).

PR #558, opened 2020-03-12
avoid OutOfBoundException when the output from delft is truncated

lfoppiano:
This patch makes grobid more robust when the delft sequences are truncated to a size lower than the input. If this happens, the output labels are defaulted to `<other>`.

See discussion on https://github.com/kermitt2/delft/pull/97


grobid-core/src/main/java/org/grobid/core/jni/DeLFTModel.java
```diff
@@ -296,11 +305,11 @@ public void run() {
...
-            label = "<other>";
-        } else if (label.startsWith("B-")) {
-            label = label.replace("B-", "I-");
-        } else if (label.startsWith("I-")) {
-            label = label.replace("I-", "");
+            label = TaggingLabels.OTHER_LABEL;
+        } else if (label.startsWith(GenericTaggerUtils.START_ENTITY_LABEL_PREFIX_ALTERNATIVE)) {
+            label = label.replace(GenericTaggerUtils.START_ENTITY_LABEL_PREFIX_ALTERNATIVE, GenericTaggerUtils.START_ENTITY_LABEL_PREFIX);
+        } else if (label.startsWith(GenericTaggerUtils.START_ENTITY_LABEL_PREFIX)) {
+            label = label.replace(GenericTaggerUtils.START_ENTITY_LABEL_PREFIX, "");
```
kermitt2: I would keep the original IOB strings and not use this `GenericTaggerUtils` stuff because it is unreadable. There is here something confusing about which annotation scheme is used, because actually it mixes here a Grobid prefix `START_ENTITY_LABEL_PREFIX` for testing a IOB prefix which is not the start of an entity but _inside_.

kermitt2: As an alternative we could introduce clearly the annotation "scheme" which is used in the constant names, e.g. `TaggingLabels.OTHER_LABEL_IOB`, `TaggingLabels.OTHER_LABEL_GROBID`, `TaggingLabels.START_ENTITY_LABEL_PREFIX_IOB`, `TaggingLabels.START_ENTITY_LABEL_PREFIX_GROBID`, etc. but I am not sure it helps in term of readability.

lfoppiano: Indeed, this is not readable. 

I've pushed a change that should make more sense. Let me know what you think

grobid-core/src/main/java/org/grobid/core/engines/label/TaggingLabels.java
```diff
@@ -15,14 +15,28 @@
 
     protected static final ConcurrentMap<Pair<GrobidModel, String>, TaggingLabel> cache = new ConcurrentHashMap<>();
 
+    //IOB labels and prefixes
+    public static final String START_ENTITY_LABEL_PREFIX_IOB_BEGINNING = "B-";
+    public static final String START_ENTITY_LABEL_PREFIX_IOB_INSIDE = "I-";
```
kermitt2: Naming confusing because of `START_ENTITY` and `INSIDE` or`START_ENTITY` plus redundant `BEGINNING` . We could have:

```
public static final String IOB_START_ENTITY_LABEL_PREFIX = "B-";
public static final String IOB_INSIDE_ENTITY_LABEL_PREFIX = "I-";
public static final String IOB_OTHER_LABEL = "O";

public static final String GROBID_START_ENTITY_LABEL_PREFIX = "I-";
public static final String GROBID_INSIDE_ENTITY_LABEL_PREFIX = "";
public final static String GROBID_OTHER_LABEL = "<other>";

public static final String ENAMEX_START_ENTITY_LABEL_PREFIX = "E-";
```

lfoppiano: Thanks! I pushed the change as requested 😉

grobid-core/src/main/java/org/grobid/core/engines/tagging/GenericTaggerUtils.java
```diff
@@ -73,8 +79,8 @@ public static String getPlainLabel(String label) {
     }
 
     public static boolean isBeginningOfEntity(String label) {
-        return StringUtils.startsWith(label, START_ENTITY_LABEL_PREFIX)
-            || StringUtil.startsWith(label, START_ENTITY_LABEL_PREFIX_ALTERNATIVE)
-            || StringUtil.startsWith(label, START_ENTITY_LABEL_PREFIX_ALTERNATIVE_2);
+        return StringUtils.startsWith(label, TaggingLabels.GROBID_START_ENTITY_LABEL_PREFIX)
+            || StringUtil.startsWith(label, TaggingLabels.IOB_START_ENTITY_LABEL_PREFIX)
+            || StringUtil.startsWith(label, TaggingLabels.ENAMEX_START_ENTITY_LABEL_PREFIX);
```
kermitt2: This method will fail with `IOB_INSIDE_LABEL_PREFIX` returning `true` for `isBeginningOfEntity()` while it is not the beginning of an entity. Problem  `IOB_INSIDE_LABEL_PREFIX` == `GROBID_START_ENTITY_LABEL_PREFIX` == "I-".
Normally this method should not be called with IOB labels... but not 100% sure.

grobid-core/src/main/java/org/grobid/core/engines/label/TaggingLabels.java
```diff
@@ -114,7 +130,7 @@
     public static final TaggingLabel EQUATION_MARKER = new TaggingLabelImpl(GrobidModels.FULLTEXT, EQUATION_MARKER_LABEL);
     public static final TaggingLabel PARAGRAPH = new TaggingLabelImpl(GrobidModels.FULLTEXT, PARAGRAPH_LABEL);
     public static final TaggingLabel ITEM = new TaggingLabelImpl(GrobidModels.FULLTEXT, ITEM_LABEL);
-    public static final TaggingLabel OTHER = new TaggingLabelImpl(GrobidModels.FULLTEXT, OTHER_LABEL);
+    public static final TaggingLabel OTHER = new TaggingLabelImpl(GrobidModels.FULLTEXT, GROBID_OTHER_LABEL);
```
kermitt2: mmm I think we should keep `OTHER_LABEL` here because all these labels are in the Grobid realm. This is confusing to add GROBID specifically here because it cannot be something else than grobid labels.

grobid-core/src/main/java/org/grobid/core/jni/DeLFTModel.java
```diff
@@ -296,11 +301,11 @@ public void run() {
 
     private static String delft2grobidLabel(String label) {
         if (label.equals("O")) {
```
kermitt2:
```java
if (label.equals(TaggingLabels.IOB_OTHER_LABEL)) {
```

grobid-core/src/main/java/org/grobid/core/engines/tagging/GenericTaggerUtils.java
lfoppiano: Yes, it has been used for grobid-ner and uses only on Grobid labels, for what I can remember. However, we might want to 
 - split the two into Grobid and IOB 
 - create a new method just for IOB (<- My preference, since by default we are talking about grobid stuff?)

grobid-core/src/main/java/org/grobid/core/engines/label/TaggingLabels.java
lfoppiano: OK, indeed!

grobid-core/src/main/java/org/grobid/core/engines/tagging/GenericTaggerUtils.java
lfoppiano: Something like: 
```
    public static String getPlainIOBLabel(String label) {
        return isBeginningOfEntity(label) ? StringUtil.substring(label, 2) : label;
    }

    public static boolean isBeginningOfIOBEntity(String label) {
        return StringUtil.startsWith(label, TaggingLabels.IOB_START_ENTITY_LABEL_PREFIX)
            || StringUtil.startsWith(label, TaggingLabels.ENAMEX_START_ENTITY_LABEL_PREFIX);
    }

    // I-<citation> --> <citation>
    // <citation> --> <citation>
    public static String getPlainLabel(String label) {
        return isBeginningOfEntity(label) ? StringUtil.substring(label, 2) : label;
    }

    public static boolean isBeginningOfEntity(String label) {
        return StringUtils.startsWith(label, TaggingLabels.GROBID_START_ENTITY_LABEL_PREFIX)
            || StringUtil.startsWith(label, TaggingLabels.ENAMEX_START_ENTITY_LABEL_PREFIX);
    }
```

lfoppiano: In `grobid-ner` there is a duplication of the same method actually

lfoppiano:
I had to correct some tests, especially related to NER, which uses `B-` as beginning, instead of `I-`. We should check grobid-ner as well and uniform everything related to grobid as `I-`.

PR #562, opened 2020-03-19
Check whether annotation is orcid and add orcid to author.

### LINKED ISSUE #512: Extraneous string "ID" added to surname with PDF and authors with ORCID links
`Hi,`
We have noticed that when processing a PDF with ORCID URL links labeled "ID", the "ID" string is concatenated to the author string.

Steps to reproduce:
For instance with this PDF:
https://journals.plos.org/plosbiology/article/file?id=10.1371/journal.pbio.3000097&type=printable

Article page: 
https://journals.plos.org/plosbiology/article?id=10.1371/journal.pbio.3000097

Using process header document + Consolidate header=true (but we have the same result with Consolidate header=false)

We get : 
```
<persName xmlns="http://www.tei-c.org/ns/1.0">
    <forename type="first">Thomas</forename>
    <surname>Burgoyne Id</surname>
</persName>
```

Whereas the expected string is  `<surname>Burgoyne</surname>`

We are using application version="0.5.6". It also happened with previous versions.

Thank you for the new release 0.5.6 and all the work.

Aazhar:
Should fix #512 
Checks whether an orcid annotation is intersecting with author names , when the annotation is pointing to an ORCID pattern then recover the author name and add ORCID ID


kermitt2:
Sorry for the slow review! It works very well when orcid annotations are present, both for correcting names and extracting the orcid and no effect on the normal cases according ot my tests. Thanks a lot Achraf !

kermitt2:
ok for some unknown reason, it does not work after merging :/

kermitt2:
OK I found the problem, it works again as expected when merged. 
I also fixed the case where `pdfAnnotation.getDestination()` was null/empty - the pdf processing was failing in this case.

PR #563, opened 2020-03-25
keep the extracted affiliations if none found from consolidation.

Aazhar:


grobid-core/src/main/java/org/grobid/core/data/BiblioItem.java
```diff
@@ -4182,27 +4182,20 @@ else if (bibo.getFullAuthors().size() == 1) {
...
-                                        aut.setSuffix(aut2.getSuffix());
-                                    break;
-                                } else if ( StringUtils.isNotBlank(aut.getFirstName()) && 
-                                    StringUtils.isNotBlank(aut2.getFirstName()) &&
-                                    (aut.getFirstName().length() == 1) && 
-                                    (aut.getFirstName().equals(aut2.getFirstName().substring(0,1))) ) {
-                                    // we have a match (initial)
+                                if ( StringUtils.isNotBlank(aut2.getFirstName()) && StringUtils.isNotBlank(aut.getFirstName())
+                                    &&( aut.getFirstName().equals(aut2.getFirstName()) ||
+                                    ((aut.getFirstName().length() == 1) && (aut.getFirstName().equals(aut2.getFirstName().substring(0,1)))))) {
```
lfoppiano, line 4187: what about the initial is with a dot? so it has a length of 2? 
What about the initial is present in the `aut2` instead of `aut`? shouldnt you check that condition too?

Aazhar: yes actually we should compare the initials for the firstnames  either for `aut2` or `aut` , WDYT ?

lfoppiano: the check (whch was like this before, so I might be wrong) first check if `aut` has length 1, and if so it compares `aut` with the first character of `aut2`

what if `aut` does not have length 1, but `aut2` is just the initial?

Aazhar: yes I didn't change this check actually, it was like this..
basically if it is  not the same firstname or `aut` firstname has more than one character then it 's not considered the same author form

Aazhar: what I suggest is  for firstnames to compare only the first char (initials)

lfoppiano: OK

I noticed another thing, at line 4205, there is `bib.setFullAuthors(bibo.getFullAuthors());` so the authors are copied anyway from the second biblio... or should I better be in bed?

lfoppiano: I added a couple of tests, there is an ignored one that is failing and I'm wondering what is the rule here

Aazhar: yes the data about authors coming from consolidation is considered more accurate than the parsed one, 
in this PR I would like to keep the affiliations if none fetched from the consolidation which seems to be more frequent

Aazhar: regarding the test, so far the affiliations from crossref are just raw and not deserialised, so maybe we could keep the affiliations parsed from the pdf ? I'm not sure how good are the affiliations from crossref actually

kermitt2:
So as explained on Mattermost, I updated this part with the new Person object deduplication, and this part becomes simply now: 

```java
       if (bibo.getFullAuthors() != null) {
            if (CollectionUtils.isEmpty(bib.getFullAuthors()))
                bib.setFullAuthors(bibo.getFullAuthors());
            else {
                // we have the complete list of authors so we can take them from the second
                // biblio item and merge some possible extra from the first when a match is 
                // reliable
                List<Person> thePersons = bib.getFullAuthors();
                thePersons.addAll(bibo.getFullAuthors());
                Person.deduplicate(thePersons);
            }
        }
```

and the deduplication handles the merging of affiliation.

kermitt2:
With this approach, it works fine when authors are well recognized initially, but when the authors are badly extracted from the PDF, we will have as authors at the same time the erroneous extraction and the correct consolidated ones. So maybe this could be refined to prioritize the consolidated authors?

kermitt2:
ok too many authors added with this approach, I'll try to find something safer...

PR #565, opened 2020-04-09
Remove side-effects when running several proceses of n-fold cross-validation of the same model while sharing the same grobid-home directory

lfoppiano:
This PR allows to run the n-fold cross-validation of a specific model several times sharing the same grobid-home. 

Currently the models are named `model_name_nfold_x` with model_name corresponding the name of the model and x the n-ith fold. However if the same model is evaluated at the same time, they will write/overwrite the models in the `grobid-home/tmp` directory. 

This PR: 
 - add a unique random string on the ith-fold model name
 - collect generated files paths and deleted them at the end of the process


kermitt2:
Thanks Luca, tested and it works fine.

Issue #531, opened 2020-01-08
header training data features columns


lfoppiano:
Following https://github.com/kermitt2/delft/issues/81 I've checked the header data and this is the result. 
Most of the header files have 32 columns, except: 

```
"31": [
        "/Users/lfoppiano/development/projects/grobid/grobid-trainer/resources/dataset/header/corpus/headers/ana0064-0199.header",
        "/Users/lfoppiano/development/projects/grobid/grobid-trainer/resources/dataset/header/corpus/headers/nihms352394.header",
        "/Users/lfoppiano/development/projects/grobid/grobid-trainer/resources/dataset/header/corpus/headers/13142_2011_Article_25.header",
        "/Users/lfoppiano/development/projects/grobid/grobid-trainer/resources/dataset/header/corpus/headers/11606_2011_Article_1868.header",
        "/Users/lfoppiano/development/projects/grobid/grobid-trainer/resources/dataset/header/corpus/headers/zpe955.header",
        "/Users/lfoppiano/development/projects/grobid/grobid-trainer/resources/dataset/header/corpus/headers/nihms366172.header",
        "/Users/lfoppiano/development/projects/grobid/grobid-trainer/resources/dataset/header/corpus/headers/12079_2011_Article_155.header",
        "/Users/lfoppiano/development/projects/grobid/grobid-trainer/resources/dataset/header/corpus/headers/11606_2011_Article_1801.header",
        "/Users/lfoppiano/development/projects/grobid/grobid-trainer/resources/dataset/header/corpus/headers/pmed.1000352.header",
        "/Users/lfoppiano/development/projects/grobid/grobid-trainer/resources/dataset/header/corpus/headers/PIIS0272638609014450.header",
        "/Users/lfoppiano/development/projects/grobid/grobid-trainer/resources/dataset/header/corpus/headers/1475-2832-3-13.header",
        "/Users/lfoppiano/development/projects/grobid/grobid-trainer/resources/dataset/header/corpus/headers/1751-0473-5-9.header",
        "/Users/lfoppiano/development/projects/grobid/grobid-trainer/resources/dataset/header/corpus/headers/onc1292.header",
        "/Users/lfoppiano/development/projects/grobid/grobid-trainer/resources/dataset/header/corpus/headers/ar900093g.header",
        "/Users/lfoppiano/development/projects/grobid/grobid-trainer/resources/dataset/header/corpus/headers/34._98-34.header",
[...]

Issue #491, opened 2019-08-24
Annotation Questions on Header


sampanriver:
I don't see any annotation guidelines for header. Could anyone give me any hint on how to annotate header?
In addition, every Chinese scientific paper has title, author, abstract and affiliations in both Chinese and English. An example is attached.
[down.pdf](https://github.com/kermitt2/grobid/files/3536948/down.pdf)
Could anyone help to tell me what should be done to deal with this situation?

kermitt2:
Hi @sampanriver ! Indeed there is no annotation guidelines for header, the reason is that we want to update the header model. It is the oldest one and it has many design flaws, it is not consistent with the principles of the other models in particular for the annotations. 

The objective to update the header model is now quite old (see #136) but there's some progress (PR #457) and an improved reading order has been added to pdfalto (but still to be more tested) to address current problem with the header. 

For this reason, there is no annotation guidelines for the header and I would not advice to produce new training data for the header before we move to the new design, because the training data might need some revision (different token order, different labels, different labelling principles). 

In the new approach (which is the one used for all the other models), only the minimum useful information is annotated and all the rest is ignored. In the current outdated header model approach, all the header content is annotated, with all the useless syntactic sugar and noise, which complexifies at the same time the annotation, the learning and the cleaning of ML results for no benefit (this approach comes from the CORA dataset which has been used when starting this project). 

In addition we want to support the new cases that you mention: multiple language version for all the fields, this is very frequent for the titles and abstracts, but as you are showing for other fields too, and additional ones like discontinuous header part more distributed in the document (thanks a lot for the testing case!). 

My answer is likely not very satisfying for you because it will certainly take a few months to make real progress on the new header model, but the cases you mention are simply not supported for the moment and the best is to make a new iteration with a more complete redesign than simply adding new labels.

kermitt2:
The header model has been rewritten, with new labels, new features, and so on. I removed all the old training data (including old CORA stuff). 

Annotation guidelines for the header model are now available: 
https://grobid.readthedocs.io/en/latest/training/header/

Issue #430, opened 2019-05-06
train title and abstract only results in poor abstract extraction


de-code:
Hi Patrice,

I am re-training the header model currently with the title and abstract only.
It does seem to work fine with the title. But the abstract extraction afterwards is very poor. It often returns authors for example. I've trained it with 100 and 1000 documents using GROBID 0.5.2 (will try 0.5.4 but it had abstract regressions). I've actually then ran the model using 0.5.3. (I will aim to keep the version consistent in the future)

My process looks roughly like:
- generate raw training data
- remove all tagging (except for `lb`)
- find title and abstract and add tags (auto-annotated tei)
- train model
- evaluate:
  - auto-annotated tei
  - grobid model on training pdf
  - grobid model on validation/test pdf

For the evaluation on auto-annotated tei I am getting a f1-score between 0.85 and 0.90 (levenshtein). Which shows that while not perfect, the training data is not that bad (some documents may not have header data).

Evaluation of the grobid model on the training pdf then ends up far below 0.10 and looking at samples it's not even close. The default model is much better.

The auto-annotated training TEI XML would look something like this:

```xml
<tei><teiHeader>
                <fileDesc xml:id="..."/>
        </teiHeader>
        <text><front><lb/>


        <docTitle><titlePart>Title line 1<lb/> and line 2<lb/></titlePart></docTitle>

        Author One and Author Two<lb/> 1<lb/>

        ABSTRACT<lb/> <div type="abstract">A long and informative<lb/> abstract...</div>.

        </front></text></tei>
```

Is that something that would get solved by using the Clusteror? Am I doing something wrong?

de-code:
I've ran a few more tests.

I was doing the training with 0.5.3 throughout and 0.5.4. For 50, 100, 200 and 500 training samples (not 1000 yet because it takes so long). I am evaluating it on 100 validation samples which could be larger but are randomly selected (but fixed, i.e. it won't change per run or future runs).

In both cases it trained the title alright. Although still below the default model. Slightly better in 0.5.4.

The abstract training really doesn't seem to work in 0.5.3 as described in the issue above (with a f1 score of far below 0.1 - in fact it was best with just 50 training samples).

Training 0.5.4 was better with a f1 score below 0.5. There is no sign of it improving with more training samples. In fact using 500 training samples yielded the worst result. As with 0.5.3, the evaluation results on training samples themselves are not much different than the validation samples (in some instances even worse).

de-code:
[...]

Issue #281, opened 2018-01-29
Different header results using processHeader or processFulltext


Aazhar:
Processing document DOI 10.1007/s10676-016-9400-6 gives different output when extracting only header metadata (poor metadata, no abstract..) comparing to extracting the fulltext.

Process Fulltext Document:

    <teiHeader xml:lang="en">
        <encodingDesc>
            <appInfo>
                <application version="0.5.1-SNAPSHOT" ident="GROBID" when="2018-01-29T13:38+0000">
                    <ref target="https://github.com/kermitt2/grobid">GROBID - A machine learning software for extracting information from scholarly documents</ref>
                </application>
            </appInfo>
        </encodingDesc>
        <fileDesc>
            <titleStmt>
                <title level="a" type="main">AI assisted ethics</title>
            </titleStmt>
            <publicationStmt>
                <publisher>Springer Nature</publisher>
                <availability status="unknown">
                    <p>Copyright Springer Nature</p>
                </availability>
                <date type="published" when="2016">2016</date>
            </publicationStmt>
            <sourceDesc>
                <biblStruct>
                    <analytic>
                        <author>
                            <persName
                                xmlns="http://www.tei-c.org/ns/1.0">
                                <forename type="first">Amitai</forename>
                                <surname>Etzioni</surname>
                            </persName>
                        </author>
                        <author>
                            <persName
                                xmlns="http://www.tei-c.org/ns/1.0">
                                <forename type="first">Oren</forename>
                                <surname>Etzioni</surname>
                            </persName>
                        </author>
                        <title level="a" type="main">AI assisted ethics</title>
                    </analytic>
                    <monogr>
                        <title level="j" type="main">Ethics and Information Technology</title>
                        <title level="j" type="abbrev">Ethics Inf Technol</title>
                        <idno type="ISSN">1388-1957</idno>
                        <idno type="eISSN">1572-8439</idno>
                        <imprint>
                            <publisher>Springer Nature</publisher>
                            <biblScope unit="volume">18</biblScope>
                            <biblScope unit="issue">2</biblScope>
                            <biblScope unit="page" from="149" to="156"/>
                            <date type="published" when="2016">2016</date>
                        </imprint>
                    </monogr>
                    <idno type="DOI">10.1007/s10676-016-9400-6</idno>
[...]

PR #580, opened 2020-05-10
[WIP] Full update of the header model

kermitt2:
Here is a largely reviewed header model:
- use of clusteror to extract labelled results, better alignment
- removal of lot's of ugly old stuff (mostly not used actually)
- discard heuristics (but still usable)
- review of features (fix old features and start adding some new ones related to font size and spacing)
- new training format (consistent with the other models)
- adaptation of the xml parser for the new training data format
- adaptation of the creation of training data for the header model
- bootstrap and manual annotation of ~300 examples of headers in the new format 
- annotation guidelines for the training data in the new format

These improvements fix or make obsolete in particular:
- duplicated introduction/abstract
- some missing text at the beginning of the body sections
- robustness with "noisy" content
- quite a lot of issues like #281, #275, #136, #128, #136, #430, #531, #491

This makes the header model consistent with the other models in term of annotation approach and should eventually improve the labelling accuracy and the quality of extracted metadata. 

The old training data (around 2600 examples) is entirely dropped. 300 examples have been annotated in the new training format with the actual header parts from the segmentation model. As compared to the old model, this provides end-to-end results for PMC already quite similar for title, keywords, and first author, but lower for full authors (authors are more duplicated with the new approach and need a deduplication) and for abstract - this is encouraging. Usage of the heuristics now degrades the end-to-end result by 2 point in f-score in average. 

To do:
- more training data (including for the segmentation model, because the header parts now always come from the segmentation model), at least to reach the old model accuracy
- deduplicate extracted authors and affiliations 
- add in the result TEI some extracted fields currently ignored (document type, journal title, group/collaboration, ...)
- remove some labels not used anymore


kermitt2:
There are now 500 annotated header examples (to be reviewed and a bit completed), including only around 30-35 PMC articles for the moment. But we're already getting better average results than the old header model (with more than 2700 annotated examples) for the end-to-end results with PMC set (f-score+0.5, with good learning curve). 
Deduplication of authors but not yet of affiliations.

PR #585, opened 2020-05-23
Try to fix the synchronisation of the language detection singleton in…

lfoppiano:
…itialisation in case of multithread environment.

There are many different theories around this, but basically what I understood is that 
 - double null check + synchronize + volatile correctly works, so thread are locked after null checking the instance... but a secondary check avoid threads that have passed the null check to repeat the initialisation. 
 - not sure it's faster than just synchronize the method every time

### DIFF
--- a/grobid-core/src/main/java/org/grobid/core/data/BiblioItem.java
+++ b/grobid-core/src/main/java/org/grobid/core/data/BiblioItem.java
@@ -42,7 +42,7 @@ import org.slf4j.LoggerFactory;
 public class BiblioItem {
     protected static final Logger LOGGER = LoggerFactory.getLogger(BiblioItem.class);
 
-    LanguageUtilities languageUtilities = LanguageUtilities.getInstance();
+    private LanguageUtilities languageUtilities = LanguageUtilities.getInstance();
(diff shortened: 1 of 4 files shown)


de-code:
But whether it is worth the time saving over just `synchronized` I don't know. How often will it be called?

Looks to be the correct way of implementing [double checked locking](https://en.wikipedia.org/wiki/Double-checked_locking).

There are actually alternatives that you could consider as well:
 [Initialization on-demand holder](https://en.wikipedia.org/wiki/Initialization_on_demand_holder_idiom) (using the classloader)

I was previously in favour of using `AtomicReference` but it would just wrap the `volatile`. I am surprised that Java still doesn't seem to offer a simple lazy initialization wrapper where you can pass in a callable (like `ThreadLocal`).

Apache commons has [LazyInitializer](https://commons.apache.org/proper/commons-lang/javadocs/api-release/org/apache/commons/lang3/concurrent/LazyInitializer.html). Maybe that would be more readable.

kermitt2:
Thank you Luca, it looks all good and solved the problem when testing with the grobid python client. Thank you @de-code for the pointers! 
So, good to learn, lazyness requires more work :D

PR #597, opened 2020-06-22
Update to gradle 6.5.1 to support JDK 13 and 14

lfoppiano:
- [x] Update to gradle 6.5
- [x] remove duplicated build code from `build.gradle` 
- [x] fix problems due to jacoco and coveralls 
- [x] fix fatJar build 
- [x] gradle install on local maven repository
- [x] fix incompatibilities between shadowJar and Gradle 6.5
- ~~[ ] bintray publication~~ can't really test it


lfoppiano:
Tested with JDK 13 and 14 (AdoptJDK) on Mac.

kermitt2:
`./gradlew clean install` gives the following messages:

```
Configuration(s) specified but the install task does not exist in project :grobid-core.
Configuration(s) specified but the install task does not exist in project :grobid-home.
Configuration(s) specified but the install task does not exist in project :grobid-service.
Configuration(s) specified but the install task does not exist in project :grobid-trainer.
```

I don't see impact actually (maybe it was present but silent before? Or maybe due to the JacocoReport task applying to the subprojects now?

lfoppiano:
ahhh, this is due to the fact that the maven plugin has been deprecated, so it was replaced by the `maven-publish` plugin which seems to be implementing only the publishing part of it. I guess this also removes the `install` task. 
I didn't test it because I stopped using the `install` task in favour of `./gradlew clean build`. 

Let me dig into it. 👍

lfoppiano:
My last commit should have fixed it. I added an additional task which is to test the bintray upload (since the publication of the maven artifact has changed, we might have some problem when publishing on bintray, and we don't want to solve them while doing the release :-) )

kermitt2:
Thanks !

Still the same message for the install task:

```
lopez@work:~/grobid$ ./gradlew clean install
Configuration(s) specified but the install task does not exist in project :grobid-core.
Configuration(s) specified but the install task does not exist in project :grobid-home.
Configuration(s) specified but the install task does not exist in project :grobid-service.
Configuration(s) specified but the install task does not exist in project :grobid-trainer.
```

It's just a warning without impact (there is no independent install at subproject level). 

So the following on line 62 seems just ignored: 

> task install(dependsOn: publishToMavenLocal)

Removing this line and adding `apply plugin: 'maven'` fixes the problem as you said, but since it's deprecated... I have no clue why the install task as added above is ignored.

lfoppiano:
I think that message is perhaps happening because the tasks are not attached immediately? 

I've tested by removing the jars from my `~/.m2/repository` directory and it seems to work. 

```
(base) Johan:grobid-source lfoppiano$ ls  ~/.m2/repository/org/grobid
grobid-home		grobid-ner		grobid-quantities
(base) Johan:grobid-source lfoppiano$ rm -rf  ~/.m2/repository/org/grobid/grobid-home/
(base) Johan:grobid-source lfoppiano$ ./gradlew clean install 
<-------------> 0% INITIALIZING [416ms]
> Evaluating settings
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by org.codehaus.groovy.reflection.CachedClass (file:/Users/lfoppiano/.gradle/wrapper/dists/gradle-6.4.1-all/13imxtezgn9nwzqt8rgtkunh1/gradle-6.4.1/lib/groovy-all-1.3-2.5.10.jar) to method java.lang.Object.finalize()
WARNING: Please consider reporting this to the maintainers of org.codehaus.groovy.reflection.CachedClass
Configuration(s) specified but the install task does not exist in project :grobid-core.
Configuration(s) specified but the install task does not exist in project :grobid-home.
Configuration(s) specified but the install task does not exist in project :grobid-service.
Configuration(s) specified but the install task does not exist in project :grobid-trainer.

> Task :grobid-core:compileJava
Note: Some input files use or override a deprecated API.
Note: Recompile with -Xlint:deprecation for details.
Note: Some input files use unchecked or unsafe operations.
Note: Recompile with -Xlint:unchecked for details.

> Task :grobid-trainer:compileJava
Note: Some input files use or override a deprecated API.
Note: Recompile with -Xlint:deprecation for details.
Note: Some input files use unchecked or unsafe operations.
Note: Recompile with -Xlint:unchecked for details.

BUILD SUCCESSFUL in 14s
27 actionable tasks: 26 executed, 1 up-to-date
(base) Johan:grobid-source lfoppiano$ ls ~/.m2/repository/org/grobid/
grobid-core		grobid-ner		grobid-service
grobid-home		grobid-quantities	grobid-trainer
(base) Johan:grobid-source lfoppiano$ 
```

kermitt2:
I was using gradle `6.5` (see my small review message). It's working indeed for the local repo install, no problem. The install task is then well executed, so I think it's just useless warning messages (using `--debug` is not helpful unfortunately).

lfoppiano:
I did not see the review message actually.. I find reviews on github hard to use. 
BTW I had a commit to update to gradle but forgot it in my local ^_^

kermitt2:
Shall we merge the branch anyway with these useless warning messages ? I don't see how to fix that.

lfoppiano:
I think the install is fine, we need to see if the bintray needs to be touched

lfoppiano:
There is another small thing I just found out. 

I think the proper way to run grobid would be to execute the jar file, instead of via Gradle. Gradle is nice for development but not for a production service, plus the horrible logs that output when it crashes it's polluting our GitHub issues list :-) 

Maybe we should add back the zip distribution or the shadow fat jar for the grobid-service. We could find a way to disable it, for example when we're running the `run` task so that the build will be faster. 

What do you think?

lfoppiano:
I found another problem with submodules, to be double-checked. 

```
com.google.inject.CreationException: Unable to create injector, see the following errors:

1) Error injecting constructor, java.lang.NoClassDefFoundError: fr/limsi/wapiti/Wapiti
  at org.grobid.core.engines.SuperconductorsParser.<init>(SuperconductorsParser.java:60)
  at org.grobid.service.SuperconductorsServiceModule.configure(SuperconductorsServiceModule.java:32) (via modules: com.google.inject.util.Modules$CombinedModule -> org.grobid.service.SuperconductorsServiceModule)
  while locating org.grobid.core.engines.SuperconductorsParser
Caused by: java.lang.NoClassDefFoundError: fr/limsi/wapiti/Wapiti
```

I tried to modify line 94 of `gradle.build` from `implementation` to `api` so that these libraries are available as transient dependencies, but it did not work.

kermitt2:
> I tried to modify line 94 of gradle.build from implementation to api so that these libraries are available as transient dependencies, but it did not work.

as transitive dependencies yes, I got my other modules failing for the same reason. 
shadowJar task was also not adding to the one-jar, because it was still referring to the dependency configuration `compile`. 

I commit a fix with an explicit copy to the jar(s), it works for me.

```shell
lopez@work:~/grobid$ unzip -l  grobid-core/build/libs/grobid-core-0.6.1-SNAPSHOT-onejar.jar  | grep wapiti
        0  2020-07-13 15:27   fr/limsi/wapiti/
      460  2014-04-29 17:41   fr/limsi/wapiti/SWIGTYPE_p_mdl_t.class
      463  2014-04-29 17:41   fr/limsi/wapiti/SWIGTYPE_p_p_char.class
      469  2014-04-29 17:41   fr/limsi/wapiti/SWIGTYPE_p_uint32_t.class
     1526  2014-04-29 17:41   fr/limsi/wapiti/Wapiti.class
     2418  2014-04-29 17:41   fr/limsi/wapiti/WapitiJNI.class
     6611  2014-04-29 17:41   fr/limsi/wapiti/opt_s.class
lopez@work:~/grobid$ unzip -l  grobid-core/build/libs/grobid-core-0.6.1-SNAPSHOT.jar  | grep wapiti
        0  2020-07-13 15:27   fr/limsi/wapiti/
      460  2014-04-29 17:41   fr/limsi/wapiti/SWIGTYPE_p_mdl_t.class
      463  2014-04-29 17:41   fr/limsi/wapiti/SWIGTYPE_p_p_char.class
      469  2014-04-29 17:41   fr/limsi/wapiti/SWIGTYPE_p_uint32_t.class
     1532  2014-04-29 17:41   fr/limsi/wapiti/Wapiti.class
     2418  2014-04-29 17:41   fr/limsi/wapiti/WapitiJNI.class
     6611  2014-04-29 17:41   fr/limsi/wapiti/opt_s.class
```

lfoppiano:
I just found an alternative solution, from here: https://github.com/johnrengelman/shadow/issues/288 

```
      from {
            project.configurations.runtimeClasspath.collect {
                it.isDirectory() ? [] : localLibs.contains(it.getName()) ? zipTree(it) : []
            }
        }

        configurations = [project.configurations.shadedLib]
        relocate 'org.apache.lucene', 'org.grobid.shaded.org.apache.lucene'
    }

    jar {
        from sourceSets.main.output

        from {
            project.configurations.runtimeClasspath.collect {
                it.isDirectory() ? [] : localLibs.contains(it.getName()) ? zipTree(it) : []
            }
        }
    }
```

kermitt2:
> I think the proper way to run grobid would be to execute the jar file, instead of via Gradle. Gradle is nice for development but not for a production service, plus the horrible logs that output when it crashes it's polluting our GitHub issues list :-)

Yes agree! I guess the shadow fat jar would do it nicely, with a bit longer build time - but the zip distribution is still useless I think and this one would take ages to build because of the grobid-home content to copy.

Then it's a pain to launch the service from a jar with a 20-kilometers command line, so we could add a launch script. like `./run.sh`?

lfoppiano:
> Yes agree! I guess the shadow fat jar would do it nicely, with a bit longer build time - but the zip distribution is still useless I think and this one would take ages to build because of the grobid-home content to copy.

Indeed. We could disable the distribution by default, and leaving the possibility to build it on demand (it's needed, for example, to build the docker image). 

> Then it's a pain to launch the service from a jar with a 20-kilometers command line, so we could add a launch script. like `./run.sh`?

Yes, the run.sh is a sweet idea indeed! :-)

kermitt2:
> @kermitt2 how big is your output jar?
-rw-rw-r-- 1 lopez lopez 17M Jul 13 15:27 grobid-core/build/libs/grobid-core-0.6.1-SNAPSHOT.jar

`runtimeClasspath` is probably better yes than making `implementation` "CanBeResolved" (it looks hacky).
I also read that for transitive dependencies we should use `api` instead of `implementation`?

Probably we don't need this for the jar task (I think it's by default always in the jar):
```
    jar {
        from sourceSets.main.output
```

kermitt2:
> Indeed. We could disable the distribution by default, and leaving the possibility to build it on demand (it's needed, for example, to build the docker image).

the Dockerfile has its own packaging instructions no? we don't a distribution task.

lfoppiano:
> > @kermitt2 how big is your output jar?
> > -rw-rw-r-- 1 lopez lopez 17M Jul 13 15:27 grobid-core/build/libs/grobid-core-0.6.1-SNAPSHOT.jar

ok same as mine :-) 

> `runtimeClasspath` is probably better yes than making `implementation` "CanBeResolved" (it looks hacky).
> I also read that for transitive dependencies we should use `api` instead of `implementation`?

Yes, it's added where needed, but we might need to flip some implementations to apis

> 
> Probably we don't need this for the jar task (I think it's by default always in the jar):
> 
> ```
>     jar {
>         from sourceSets.main.output
> ```

Yes indeed. 

I can add my modification, and leave your commented if it's ok for you.

lfoppiano:
> the Dockerfile has its own packaging instructions no? we don't a distribution task.

Yes, but it's assuming the distribution zip is there as it's unzipping and spreading in the docker container... So we could just use a custom task for that which is used only when needed (and for whom wants a distribution zip).

kermitt2:
> Yes, but it's assuming the distribution zip is there as it's unzipping and spreading in the docker container... So we could just use a custom task for that which is used only when needed (and for whom wants a distribution zip).

Ah yes indeed ! But it could be simpler and more readable/transparent to simply copy the stuff in the Dockerfile without any distribution zip for grobid-home and grobid-service (and grobid-trainer in the future, because I've developed a training REST API)? If it's used only for the docker image.

lfoppiano:
I've updated the release and bintray plugins, but I cannot test them from a branch without a real release 😅 

Everything seems to work, except running the fat jar of grobid-core: 

```
(grobidSuperconductors2) lucas-mbp:grobid lfoppiano$ java -jar grobid-core/build/libs/grobid-core-0.6.1-SNAPSHOT-onejar.jar 
Exception in thread "main" java.lang.NoClassDefFoundError: org/slf4j/LoggerFactory
	at org.grobid.core.engines.ProcessEngine.<clinit>(ProcessEngine.java:40)
	at org.grobid.core.main.batch.GrobidMain.main(GrobidMain.java:174)
Caused by: java.lang.ClassNotFoundException: org.slf4j.LoggerFactory
	at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:602)
	at java.base/jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(ClassLoaders.java:178)
	at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:522)
	... 2 more
(grobidSuperconductors2) lucas-mbp:grobid lfoppiano$ 
```

lfoppiano:
OK. Commit 1098177 fixes the issue reported at https://github.com/kermitt2/grobid/pull/597#issuecomment-659838850

Issue #529, opened 2019-12-24
Update Jep to 3.9.0


lfoppiano:
As in the title ;-) might have improved some of the glitches with the previous version. 

https://github.com/ninia/jep/releases/tag/v3.9.0

Although I could not find the latest changes or a release notel...

lfoppiano:
With jdk 13 and 14, we've got the following issue: 
```
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release
INFO  [2020-06-30 04:04:45,306] com.hubspot.dropwizard.guicier.DropwizardModule: Added guice injected health check: org.grobid.service.resources.HealthResource
INFO  [2020-06-30 04:04:45,311] org.grobid.core.utilities.GrobidProperties: environment variables overrides: {}
INFO  [2020-06-30 04:04:45,313] org.grobid.service.GrobidRestService: Initiating Servlet GrobidRestService
INFO  [2020-06-30 04:04:45,314] org.grobid.core.main.LibraryLoader: Loading external native sequence labelling library
INFO  [2020-06-30 04:04:45,315] org.grobid.core.main.LibraryLoader: Loading Wapiti native library...
INFO  [2020-06-30 04:04:45,316] org.grobid.core.main.LibraryLoader: Loading JEP native library for DeLFT... /Users/lfoppiano/development/projects/grobid/grobid-home/lib/mac-64
com.google.inject.CreationException: Unable to create injector, see the following errors:

1) Error injecting constructor, org.grobid.core.exceptions.GrobidException: [GENERAL] Loading JEP native library for DeLFT failed
  at org.grobid.service.GrobidRestService.<init>(GrobidRestService.java:69)
  at org.grobid.service.modules.GrobidServiceModule.configure(GrobidServiceModule.java:30) (via modules: com.google.inject.util.Modules$CombinedModule -> org.grobid.service.modules.GrobidServiceModule)
  while locating org.grobid.service.GrobidRestService

1 error
        at com.google.inject.internal.Errors.throwCreationExceptionIfErrorsExist(Errors.java:470)
        at com.google.inject.internal.InternalInjectorCreator.injectDynamically(InternalInjectorCreator.java:184)
        at com.google.inject.internal.InternalInjectorCreator.build(InternalInjectorCreator.java:110)
        at com.google.inject.Guice.createInjector(Guice.java:99)
        at com.google.inject.Guice.createInjector(Guice.java:84)
[...]

PR #603, opened 2020-07-01
[WIP] Update JEP, support for JDK >11 and general integration of python code 

lfoppiano:
- [x] update jep to 3.9.0 #529 
- [ ] remove workaround for loading the native libraries which will fail with JDK > 11
- [ ] study how to apply additional integration of python code [bonus]


lfoppiano:
This is obsolete as the implemented features are already implemented and the non-implemented features are already reported as issues.

PR #605, opened 2020-07-06
Fix the output filename wrongly taken from the input form when the text form is used

lfoppiano:
The button download tei did not work for text forms, because the filename was taken from the file upload input form


kermitt2:
Thanks !

PR #606, opened 2020-07-07
Add missing link

koppor:
KeepAChangelog requires to have links in the header. This one adds it.

grafik

Furthermore, during development, there should be an `[Unreleased]` heading.


kermitt2:
Well noted. Thank you Oliver !

PR #607, opened 2020-07-21
65x performance and memory optimisation for process citation

murkutep:
This PR intends to merge changes related to performance and memory optimisations
The details are posted below

1. pre-compile regex
   --> reduces mem usage
   --> improves performance as regex is precompiled
2. pre-init lexicon dependencies
   --> lazy initialization broke singleton in concurrent environment
   --> decreased memory footprint

We did GROBID performance profiling
Tools used:
Benchmarking tool: visualVM
Testing framework: locustio

Observations before optimisations:
memory_before_optimisation
Heap dump before optimisation. 
  
threads_before_optimisation
Thread dump before optimisation. 
  
We discovered following major issues around the code base in focus:
1. Regex Patterns were not precompiled
2. Lazy initialization of Singletons (can cause weak singleton)
3. String manipulations without StringBuilders

Observations after optimisations:
memory_after_optimisation
Heap dump after optimisation. 
  

threads_after_optimisation
Thread dump after optimisations. 
  

  | Before optimization | After optimization
-- | -- | --
GROBID performance | <ul><li>Total users: 100.</li><li>Total runtime: 600s (10 min)</li><li> Total requests: 123<li>Failures: 121</li><li> Throughput: 0.21 requests/sec</li><li> Memory consumption 4 GB </li></ul> | <ul><li> Total users: 100</li><li> Total runtime: 600s (10 min)</li><li> Total requests: 7915</li><li> Failures: 0</li><li> Throughput: 13.17 requests/sec</li><li> Memory consumption 750 MB</li></ul>

Co-authored-by: rvkhakhkhar <rvkhakhkhar@gmail.com>
Co-authored-by: imprashant <imprashantm@gmail.com>

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -217,6 +217,7 @@ project("grobid-core") {
         compile "commons-io:commons-io:2.5"
         compile "org.apache.commons:commons-lang3:3.6"
         compile "org.apache.commons:commons-collections4:4.1"
+        compile "org.apache.commons:commons-text:1.8"
         compile "commons-dbutils:commons-dbutils:1.7"
         compile "org.apache.httpcomponents:httpclient:4.5.3"
         compile "xerces:xercesImpl:2.12.0"
--- a/grobid-core/src/main/java/org/grobid/core/engines/AuthorParser.java
+++ b/grobid-core/src/main/java/org/grobid/core/engines/AuthorParser.java
@@ -1,5 +1,6 @@
 package org.grobid.core.engines;
 
+import java.util.regex.Pattern;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.collections4.CollectionUtils;
 
@@ -33,6 +34,7 @@ import java.util.StringTokenizer;
  * @author Patrice Lopez
  */
 public class AuthorParser {
+    public static final Pattern REGEX_PATTERN  = Pattern.compile("et\\.? al\\.?.*$");
 	private static Logger LOGGER = LoggerFactory.getLogger(AuthorParser.class);
     private final GenericTagger namesHeaderParser;
     private final GenericTagger namesCitationParser;
@@ -50,7 +52,7 @@ public class AuthorParser {
             return null;
         }
 
-        input = input.trim().replaceAll("et\\.? al\\.?.*$", " ");
+        input = REGEX_PATTERN.matcher(input.trim()).replaceAll(" ");
 
         // for language to English for the analyser to avoid any bad surprises
         List<LayoutToken> tokens = GrobidAnalyzer.getInstance().tokenizeWithLayoutToken(input, new Language("en", 1.0));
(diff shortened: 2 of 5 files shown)


kermitt2:
Thanks a lot for the PR and the effort to improve the processing of citations!

I did some performance tests on my side with the current master and this PR version, but did not see differences. Using jmeter and jprofiler on 100,000 processCitation queries with 100 concurrent queries, I have the following with the PR:

grobid-pr607-summary
grobid-pr607-jprofiler

And with the master version:

grobid-master-summary
grobid-master-jprofiler

So I don't see any significant difference, both for memory and runtime, which is more or less what I would expect because:
- the use of regular expressions is really negligible in comparison to the full chain of ML processing, so optimizing them could not really impact the overall runtime 
- lexicon init is done one time

These changes could not explain a difference of 65 times I think, because they touch very minor aspects of the whole process. 

So far I've not seen issues with the lazy initialization of the lexicon, the benchmarking that I am doing regularly gives the same results with one or several threads. Luca fixed the problem of the lazy init for the resources of the language recognizer a few month ago with PR #585, but I don't think it is similar for the full lexicon object. However I can change the lazy initialization of the lexicon in a similar manner (double check synchro on the instance) to be sure.

Now how to explain the 65 times difference you report? I could imagine one of the following:
- you ran the master version with consolidation option, while the PR is running without
- the evaluation of the master version was done on the server "cold". The loading of the models and lexical information takes some time and the server might have been overloaded with queries at the same time, resulting in a lot of 503 responses and some delay. But after a while, it should be on the correct track, and I doubt it could explain a 65x difference alone.
- you are using a Windows environment where the multithreading of the current master appears to be broken (we are focusing on Linux and macos support so far, given our limited availability for this project)

khakhkharr:
@kermitt2 

Thanks for responding to our PR.
Just to correct, we using MacOS environment for performance testing.

We will get get back to you with more accurate information about what data, script and configuraion we used to benchmark `api/processCitations` without consolidation option.

lfoppiano:
@murkutep @imprashant do you have more comments? May I close this PR?

pmurkute:
@lfoppiano happy to close it, thanks.

kermitt2:
I didn't close it as a reminder to add the lazy init of Lexicon (but it is added in PR #687 with other changes on Lexicon.java) and to pre-compile the regex (not done yet.... though I did not observe actual runtime change, it would be good to add). 

In PR #687 there is a a runtime gain of around 10%, but it was by improving the alignment between document layout tokens and labelled tokens returned by the sequence labelling algorithm, which is something I spotted by the breakdown of runtime per method produced by JProfiler. The total time spent on these regex is really small (less than 2% for full document processing). 

One interesting "hotspot" for runtime is however the time spent by FastMatcher (around 15% of the whole runtime), in particular matching the journal names (11.3%). There are apparently too many short abbreviated concurrent journal names matching at the same time in text body content, and this could certainly be optimized.

kermitt2:
Issue about potential optimization for FastMatcher and journal name: #731

Pre-compiled regex added in f466930dbca9ab39b4f9a50b1df260a973b6cab4 which adds a few percents speed-up following some quick tests. 

Thank you again @murkutep for the issue !

Issue #608, opened 2020-07-23
Not currently possible to configure all of grobid.properties with environment variables


bananaoomarang:
Apologies if I have missed an existing issue or piece of documentation but I think this is the case...

Currently it is possible to override any property starting with `grobid` but not the properties starting with `org`, since these are not picked up [here](https://github.com/kermitt2/grobid/blob/master/grobid-core/src/main/java/org/grobid/core/utilities/GrobidProperties.java#L342-L348).

For example the property `org.grobid.max.connections` would be the environment var `ORG__GROBID__MAX__CONNECTIONS` but when setting this I do not see the override appear in [this log](https://github.com/kermitt2/grobid/blob/master/grobid-core/src/main/java/org/grobid/core/utilities/GrobidProperties.java#L346).

However, if I set `GROBID__ORG__GROBID__MAX__CONNECTIONS` I see it show up in the log as `grobid.org.grobid.max.connections`, but this will not override the intended setting.

When deploying with Docker it is very convenient to be able to override settings on the environment so I think it would be good if all properties were able to be set.

PR #574, opened 2020-04-22
Correcting grobid.properties properties name

lfoppiano:
This should fix #571

PR #609, opened 2020-07-23
Allow overriding org.* properties with env vars

bananaoomarang:
Fixes: https://github.com/kermitt2/grobid/issues/608

This changes `EnvironmentVariableProperties` to allow passing in a regex instead of a prefix, allowing `GrobidProperties` to match both `ORG__GROBID__*` and `GROBID__` environment variables


de-code:
> Looks good.
> 
> Just a comment, I'm not sure grobid is supposed to have `org.grobid` prefixes configuration items, I remember having submitted #574 but it was never merged.

I was wondering as well, why we need both prefixes. I would be in favour of getting #574 merged instead. Then I believe we don't need this change anymore? Or is there still value having it for sub-modules?

kermitt2:
Thanks Milo !

PR #614, opened 2020-08-02
Improve callout to reference match

mash:
Improve callout to reference match by removing parentheses often seen in callouts.

As seen in the docs, callouts often include parentheses.
https://grobid.readthedocs.io/en/latest/training/fulltext/

I saw many occasions that these parentheses prevent matching the correct references.
See the red boxes when trying arxiv.org/pdf/1801.01290.pdf

Screen Shot 2020-08-02 at 12 56 29


kermitt2:
Many thanks @mash ! Very good finding, for some reasons I was thinking that the parenthesis were removed somewhere upstream in the process, but you're absolutely right. 

In addition to your example PDF, here's the evaluation for callout resolution on the PMC set (1943 PDF - 139,835 reference callouts), branch `update_header` without your PR:

```
Precision citation contexts:     81.32
Recall citation contexts:        69.13
fscore citation contexts:        74.73
```

and with your PR: 

```
Precision citation contexts:     81.41
Recall citation contexts:        69.86
fscore citation contexts:        75.2
```

Note that there are many references with numerical reference markers in PMC, so the positive impact should be much larger with arXiv for instance.

PR #620, opened 2020-08-13
NPE when Grobid is initialised from a module or test and the property file is not loaded yet (e.g. static context initialisation of GrobidModels)

lfoppiano:
When we invoke the static context of GrobidModels, all the models declared over there are loaded. Before applying #559 this was working without issues. 559 introduce a lookup on the properties that, at the time of the test is running are not loaded, therefore NPE. 

The first commit fixes right away the issue. However, I'm wondering whether we should have a completely different class for the DUMMY model, because if the test is ran from a directory that has no access to the grobid-home, I fear it will fail anyway.

### DIFF
--- a/grobid-core/src/main/java/org/grobid/core/utilities/GrobidProperties.java
+++ b/grobid-core/src/main/java/org/grobid/core/utilities/GrobidProperties.java
@@ -240,18 +240,21 @@ public class GrobidProperties {
     }
 
     /**
-     * Return the value corresponding to the property key. If this value is
-     * null, return the default value.
+     * Return the value corresponding to the property key. If the properties are not initialised, it returns null
      *
      * @param pkey the property key
      * @return the value of the property.
      */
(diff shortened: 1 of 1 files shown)


lfoppiano:
Actually, this fix seems enough, when the grobid-home is not reachable, each model is loaded using a path `my_Root/models/affiliation-address/model.wapiti` which doesn't bring any harm if the model is not used, which is not the case since this scenario happens in the scope of a unit test.

lfoppiano:
Just found out that this problem is occurring for every submodule, and not only for the DUMMY model. When the GrobidModels class is initialised and the property file not yet loaded, as looking up for the "engine configuration" for each model, it will result in NullPointerException. 

E.g. try to run the training of grobid-quantities.

lfoppiano:
Related to #559

kermitt2:
Thank you !

PR #624, opened 2020-08-19
update call to crossref when title is supplied

lfoppiano:
I've noticed that when the title is supplied, grobid uses a deprecated field `query.title` (https://github.com/CrossRef/rest-api-doc#works-field-queries) instead of `query.bibliographic`. 

This results in a 400 Bad request from crossref, and the consolidation is not applied.

### DIFF
--- a/grobid-core/src/main/java/org/grobid/core/utilities/Consolidation.java
+++ b/grobid-core/src/main/java/org/grobid/core/utilities/Consolidation.java
@@ -36,6 +36,8 @@ import java.util.HashMap;
 import com.rockymadden.stringmetric.similarity.RatcliffObershelpMetric;
 import scala.Option;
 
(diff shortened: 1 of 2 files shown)


kermitt2:
Good catch, thank you !!

lfoppiano:
oh.. sorry about that... I didn't notice the `!=` in the if, and worst of all when I tested it, as far as I remember it was working, but it does not make sense ...

kermitt2:
Actually I didn't notice that it was not working properly since 8 months :D 
So no problem, it works fine and it is limited to crossref now. The code simplification could be added anyway directly, I was a bit in a hurry so I didn't make the change.

lfoppiano:
OK. No worries for the code simplification, they are optional ;-)

PR #630, opened 2020-08-28
training data to improve prediction of authors combined with some ids (orcid)

Aazhar:
this improves the fix for #512


kermitt2:
Some test cases, more or less anything recent from PLOS:  https://journals.plos.org/plosone/article?id=10.1371/journal.pone.0245039
https://journals.plos.org/plosone/browse/biology_and_life_sciences

Issue #628, opened 2020-08-27
Docker build broken in current master


de-code:
Running the Docker build is currently broken in `master` and also in `0.6.1` (but not in `0.6.0`).

Command:

```bash
docker build .
```

Error:

```
Step 20/37 : COPY --from=builder /opt/grobid-source/grobid-core/build/libs/grobid-core-*-onejar.jar ./grobid/grobid-core-onejar.jar
 ---> f50b25031de8
Step 21/37 : COPY --from=builder /opt/grobid-source/grobid-service/build/distributions/grobid-service-*.zip ./grobid-service.zip
COPY failed: no source files were specified
```

This seems to be caused by https://github.com/kermitt2/grobid/commit/683804d31eb5635468c93195a0bf79d7f91db1ed which disabled the creation of the distribution zip.

Might be good to add the Docker build as part of the CI build.

/cc @lfoppiano

lfoppiano:
Yes, we would need to add a new task that creates the zip/tar file, which can be used for the docker, or we could modify the current docker builder to use the various jars (less clean).

lfoppiano:
The quick fix is to set `tasks.distZip.enabled` to true in the `grobid-service` section: 

```
    tasks.distZip.enabled = true
```

The correct fix is to create a new task that would do that, and call this task when building the new docker image. I've spent some time already without managing to have it work.

kermitt2:
What about the following: 

- in `gradle.build`, line 333

```
tasks.distZip.enabled = true
```

- in the docker file, line 40:

```
RUN ./gradlew distZip
```

and that's it, no distribution for the simple `./gradlew clean install`, only when building a new docker image.

```
Step 22/38 : COPY --from=builder /opt/grobid-source/grobid-service/build/distributions/grobid-service-*.zip ./grobid-service.zip
 ---> dd1f627afab9
Step 23/38 : COPY --from=builder /opt/grobid-source/grobid-home/build/distributions/grobid-home-*.zip ./grobid-home.zip
 ---> 2511627a323c
Step 24/38 : RUN unzip -o ./grobid-service.zip -d ./grobid &&     mv ./grobid/grobid-service-* ./grobid/grobid-service
 ---> Running in 04dad7d7bb09
```

kermitt2:
> Might be good to add the Docker build as part of the CI build.

This would make each CI build at 30 minutes, so degrading a lot the interest of the CI.

de-code:
[...]

PR #636, opened 2020-09-07
enabled service dist zip for docker build

de-code:
quick "fix" for https://github.com/kermitt2/grobid/issues/628

This is to make it work again. We could optimise it later.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -330,7 +330,7 @@ project(":grobid-service") {
         all*.exclude group: 'log4j', module: "log4j"
     }
 
-    tasks.distZip.enabled = false
+    tasks.distZip.enabled = true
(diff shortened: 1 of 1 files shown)


lfoppiano:
The problem with this is that the zip will be created when the "build" task is run.

de-code:
> The problem with this is that the zip will be created when the "build" task is run.

From my brief investigation, it looked easier to exclude tasks via the gradle CLI than it is to enable tasks. So maybe that could be used.
But I thought as a first step it would be good to make it work again and then try to make it faster while keeping the docker build working.
What do you think?

lfoppiano:
If @kermitt2 agrees, so do I :-)

kermitt2:
go :)

PR #648, opened 2020-10-12
Fixing problem in PR 559

lfoppiano:
I've added two new methods in the GrobidProperties that allow to: 
 - get all the engines selected individually in the configuration 
 - get all possible engines needed (result of the above point  and the engine selected in the `grobid.crf.engine` property. 

When running LibraryLoader.load() instead of checking only the value of `grobid.crf.engine` I consider the list of specified engines.

### DIFF
--- a/grobid-core/src/main/java/org/grobid/core/main/LibraryLoader.java
+++ b/grobid-core/src/main/java/org/grobid/core/main/LibraryLoader.java
@@ -1,5 +1,6 @@
 package org.grobid.core.main;
 
+import org.apache.commons.collections4.CollectionUtils;
 import org.apache.commons.lang3.ArrayUtils;
 import org.apache.commons.lang3.SystemUtils;
 import org.grobid.core.engines.tagging.GrobidCRFEngine;
@@ -11,12 +12,12 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.File;
-import java.io.FileFilter;
 import java.io.FilenameFilter;
 import java.lang.reflect.Field;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.Arrays;
+import java.util.Collections;
 
 import static org.apache.commons.lang3.ArrayUtils.isEmpty;
 
@@ -95,8 +96,8 @@ public class LibraryLoader {
 
             }
 
-            if (GrobidProperties.getGrobidCRFEngine() == GrobidCRFEngine.WAPITI ||
-                GrobidProperties.getGrobidCRFEngine() == GrobidCRFEngine.DELFT) {
+            if (CollectionUtils
+                .containsAny(GrobidProperties.getDistinctModels(), Arrays.asList(GrobidCRFEngine.WAPITI, GrobidCRFEngine.DELFT))) {
                 // note: if DeLFT is used, we still make Wapiti available for models not existing in DeLFT (currently segmentation and 
                 // fulltext)
                 File[] wapitiLibFiles = libraryFolder.listFiles(new FilenameFilter() {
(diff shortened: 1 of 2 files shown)


grobid-core/src/main/java/org/grobid/core/main/LibraryLoader.java
```diff
@@ -126,7 +129,7 @@ public boolean accept(File dir, String name) {
                     try {
                         System.load(wapitiLibFiles[0].getAbsolutePath());
                     } finally {
-                        if (GrobidProperties.getGrobidCRFEngine() == GrobidCRFEngine.DELFT) {
+                        if (CollectionUtils.containsAny(distinctModels, Arrays.asList(GrobidCRFEngine.DELFT))) {
```
de-code, line 132: I guess always using `singletonList` was too boring :D

grobid-core/src/main/java/org/grobid/core/utilities/GrobidProperties.java
```diff
@@ -312,6 +311,23 @@ private void init() {
...
+        return new HashSet(modelSpecificEngines);
+    }
+
+    /** Return the distinct values of all the engines specified in the individual model configuration in the property file **/
+    public static List<GrobidCRFEngine> getModelSpecificEngines() {
+        return getProps().keySet().stream()
+            .filter(k -> ((String) k).startsWith(GrobidPropertyKeys.PROP_GROBID_CRF_ENGINE + '.'))
+            .map(k -> GrobidCRFEngine.get(StringUtils.lowerCase(getPropertyValue((String) k))))
+            .distinct()
+            .collect(Collectors.toList());
```
de-code, line 328: perhaps use `toSet`?

grobid-core/src/main/java/org/grobid/core/utilities/GrobidProperties.java
```diff
@@ -312,6 +311,23 @@ private void init() {
         loadCrfEngine();
     }
 
+    /** Return the distinct values of all the engines that are needed */
+    public static Set<GrobidCRFEngine> getDistinctModels() {
+        List<GrobidCRFEngine> modelSpecificEngines = getModelSpecificEngines();
+        modelSpecificEngines.add(getGrobidCRFEngine());
```
de-code, line 317: minor: shame that core Java doesn't seem to have a functional union method. Could use Guava's [Sets.union](https://guava.dev/releases/18.0/api/docs/com/google/common/collect/Sets.html#union(java.util.Set,%20java.util.Set)).
Otherwise I'd perhaps prefer to create the result HashSet first and then modify it (i.e. not modify the result list from another method).

de-code:
Looking good too me. Thank you for doing that.

lfoppiano:
Thanks @de-code  for checking it out. I move it to "ready".

lfoppiano:
(I suggest to also apply #620, as it's touching a similar part)

grobid-core/src/main/java/org/grobid/core/utilities/GrobidProperties.java
lfoppiano: good point

grobid-core/src/main/java/org/grobid/core/main/LibraryLoader.java
lfoppiano: singletonList is for single is for single element's lists ;-)

lfoppiano:
@de-code Thanks for the feedback. I pushed a fix.

kermitt2:
Very readable, tested and it fixes the problem raised in PR #559, thanks a lot! 
It's much easier with this fix to specify a couple of models with DeLFT, the rest with Wapiti - which is the best config currently in term of accuracy (the 2 DeLFT models to use for this being the citation model and the affiliation-address model).

grobid-core/src/main/java/org/grobid/core/main/LibraryLoader.java
de-code: Isn't the right hand side containing just one element like above, i.e. `GrobidCRFEngine.DELFT`? I hadn't had my coffee yet though :D
(I only noticed that there is a difference, it doesn't really matter)

PR #658, opened 2020-10-18
Architecture parameter

kermitt2:
Possibility to add a parameter in the config properties file for selecting the architecture of choice for the DeLFT model (default is `BidLSTM-CRF`). For the moment only a default architecture selection for all models is considered, it would need to be extended per model similarly as the sequence labelling engine. 

```
grobid.delft.architecture=BidLSTM_CRF
```

At instantiation of the model, the model files corresponding to the architecture have to be present under `grobid-home/models/`.


kermitt2:
Looks good, I merge... Next step would be to have the architecture defined at model level, like the engine, which is straightforward,

PR #670, opened 2020-11-03
avoiding segmenting sentences with references at the beginning of the sentence

lfoppiano:
This PR adds a heuristic that allows the reposition of references falling at the beginning of a sentence, to the end of the previous one


lfoppiano:
I reverted the commits that were not related to this PR. 

Ideally it's probably better to cherry-pick just the relevant commits:  
 - 2d3c60f (implementation) , 
 - e81bdc5 (just minor rewriting), 
 - 2bfd557 (merge with other improvements done)

kermitt2:
This heuristics is already in master or I missed something?

lfoppiano:
In SentenceUtilities, the main part is at line 144, there is an additional block of code that moves the references at the beginning of the sentence, to the previous one.

kermitt2:
> In SentenceUtilities, the main part is at line 144, there is an additional block of code that moves the references at the beginning of the sentence, to the previous one.

Yes but I mean this is already done by the following block lines 169 to 279, in a more controlled and reliable manner (constrained by numerical superscript reference marker)? 

In general just moving all references at the beginning of a sentence to the previous one is not always valid, e.g.things like this are quite common:

```xml
<s><ref type="bibr">Peskun (2001)</ref> shows that the Metropolis-Hastings algorithm leads to estimates with smallest asymptotic variance in a class of variants...</s>
```

```xml
<s><ref type="bibr">(Pritchard and others 2000)</ref> subpopulations is a measure of the number of population subgroups in the data set.</s>
```

lfoppiano:
oh I see, then we can trash this PR 😅

PR #672, opened 2020-11-10
[WIP] Add more training data 

lfoppiano:
This PR provides: 
 - additional training data for full-text and segmentation
 - new models for full-text and segmentation, trained with all available data
 - end 2 end evaluation using both PMC and bioxiv

@kermitt2 At this stage, I'm looking for feedback on the added training data. We could continue to iterate over this PR before merging it until we reach a satisfying improvement. 
I'm setting this as a draft for the time being.


lfoppiano:
In fact, I just realised that by creating this training data with mac there are some missing numbers... therefore I will correct these files again

lfoppiano:
Adding the bioxriv fulltext data indeed improved the results on the e2e both bioxirv and pmc. 
Compared with the last official result from Patrice (PMC_sample_1943.results.grobid-0.6.2-SNAPSHOT-Glutton-DeLFT-WAPITI-MIXED-SciBERT-31.10.2020) we've got: 
- soft matching from 65.45 to 68.28 micro avg F1 score
- strict matching from 60.25 to 65.13 micro avg F1 score

lfoppiano:
I close this as it's obsolete

Issue #676, opened 2020-11-23
Potential NullPointerException in FullTextParser if segmentation is not resulting in body


de-code:
I encountered a `NullPointerException` when trying to process a document of the bioRxiv 10k test dataset (`262469v1`, please do not use it for training).

Unfortunately I can't produce it consistently, I believe due to https://github.com/kermitt2/pdfalto/issues/107#issue-749023503
(it is also model specific)

Due to the spacing issue I do not expect it to perform well. But it shouldn't result in a `NullPointerException`.

<details>
<summary>logs (`elifesciences/sciencebeam-trainer-delft-grobid:0.0.9` which should be mostly the same as `0.6.1`)</summary>

```text
ERROR [2020-11-18 16:40:58,583] org.grobid.core.engines.FullTextParser: DocumentPointer for block 663 points to 19378 token, but block token size is 10134
ERROR [2020-11-18 16:40:58,584] org.grobid.service.process.GrobidRestProcessFiles: An unexpected exception occurs. 
! java.lang.NullPointerException: null
! at org.grobid.core.engines.tagging.GenericTaggerUtils.processLabeledResult(GenericTaggerUtils.java:61)
! at org.grobid.core.engines.tagging.GenericTaggerUtils.getTokensWithLabelsAndFeatures(GenericTaggerUtils.java:57)
! at org.grobid.core.tokenization.TaggingTokenSynchronizer.<init>(TaggingTokenSynchronizer.java:35)
! at org.grobid.core.tokenization.TaggingTokenClusteror.<init>(TaggingTokenClusteror.java:58)
! at org.grobid.core.engines.FullTextParser.processFigures(FullTextParser.java:1903)
! at org.grobid.core.engines.FullTextParser.processing(FullTextParser.java:255)
! ... 76 common frames omitted
! Causing: org.grobid.core.exceptions.GrobidException: [GENERAL] An exception occurred while running Grobid.
! at org.grobid.core.engines.FullTextParser.processing(FullTextParser.java:312)
! at org.grobid.core.engines.FullTextParser.processing(FullTextParser.java:113)
! at org.grobid.core.engines.Engine.fullTextToTEIDoc(Engine.java:456)
! at org.grobid.core.engines.Engine.fullTextToTEI(Engine.java:447)
! at org.grobid.service.process.GrobidRestProcessFiles.processFulltextDocument(GrobidRestProcessFiles.java:181)
[...]

PR #677, opened 2020-11-24
added workaround for NPE in FullTextParser

de-code:
workaround for https://github.com/kermitt2/grobid/issues/676

probably doesn't solve some underlying issue


lfoppiano:
Might be the opportunity to refactor a bit the FulltextParser, when the string is null, we might want to skip all the processing of figures and tables directly.

de-code:
> Might be the opportunity to refactor a bit the FulltextParser, when the string is null, we might want to skip all the processing of figures and tables directly.

Yes, this is just a workaround to the issue that I encountered, but don't fully understand. e.g. is it normal that `featSeg` is not `null` but doesn't produce any text? If so, in what situation would it be `null`?

In general the function could probably do with breaking up. It might be good to get refactored code merged quickly to avoid conflicts. Perhaps that would work well for pairing. (Although not so well with major timezone differences)

lfoppiano:
I close this as the commit has been included in #684

PR #682, opened 2020-12-18
use Crossref-Plus-API-Token instead of Authorization

kermitt2:
CrossRef web API now request the metadata plus token to be put in Crossref-Plus-API-Token header:

https://github.com/CrossRef/rest-api-doc#authorization-token-for-plus-service


kermitt2:
@Aazhar nothing else right?

Aazhar:
@kermitt2 , yes that's all , thanks :)

PR #684, opened 2020-12-24
Avoid NPE in FulltextParser when the body text is empty or null

lfoppiano:
Continuation of #677 

I moved the check before labelling sequence for full-text, so that if there is no body all the subsequent chain in skipped.


kermitt2:
Thanks !

Issue #576, opened 2020-04-30
Suggestion: Binaries, Data and Models as dependencies


de-code:
I  know that nowadays there is often talk of keeping the data close to the code. But maybe best not to put it too close...

My suggestion would be to:

- Binaries (such as Wapiti, pdfalto etc)
  - Version and release binaries separately (e.g. as an attachment to a GitHub release)
  - Declare release version as a dependency (similar to maven dependencies)
  - Download version based on dependency
  - Advantages:
    - easy tracking of what version is used
    - external projects have better access to the binaries
- Models (Wapiti and DL models)
  - Version and release models separately (not sure what the best bucket would be, could also be attached to a GitHub release if it makes sense)
  - Would also suggest to gzip the models where it makes sense (for Wapiti models it does)
  - Declare release version as a dependency
  - Download model version
  - Advantages:
    - easier tracking of model version (and hopefully with it where it came from)
    - only models that are used would get downloaded (e.g. not DL when actually Wapiti is used)
- Training data
  - It definitely makes sense to keep the training data in Git but I would suggest in a separate data repository
  - Data can be version or simply by pointing to the git hash
  - Releasing zipped versions probably makes sense (and training could unzip on the fly)
  - Advantages:
    - the reference could be included in the model meta data, helping to identify what data was used for training
    - developers not re-training models or not all of them don't need to download all of the data

The general advantages would be to keep the git repository and the git repository download small. Especially the repository will increase as it is keeping all of the history.

Current git repository is about 1.9 GB after cloning, 275 MB when downloaded as a zip (without history).

I would also apply the same suggestion to the DeLFT repository.

For example if another Python repository where to depend on the github version (because it might include features not yet released), then it would have to at least download the zip. For a Java projects it may be less common but still similar.

Potential alternatives might be git submodules or git lfs. Although I think I am not in favour of those.

/cc @kermitt2

PR #687, opened 2021-01-01
New yaml config

kermitt2:
This PR introduces a new configuration using a yaml file, instead of properties. There is now only one config file (server has not anymore its own). See [grobid.yaml](https://github.com/kermitt2/grobid/blob/e7c08d05c44d1b6d82f8088d374e98216a341f0e/grobid-home/config/grobid.yaml)

Custom config can be easily passed to run a docker image without env. variables:
```console
docker run --rm --gpus all --init -v /home/lopez/grobid/grobid-home/config/grobid.yaml:/opt/grobid/grobid-home/config/grobid.yaml:ro  grobid/grobid:0.6.2-SNAPSHOT
```

Notable change:
- each model has now its own parameters (engine, DL archtecture, training parameters, ...), we don't have any more the confusing default parameters or only general-level parameters:
```yaml
   - name: "citation"
      engine: "wapiti"
      #engine: "delft"
      wapiti:
        # wapiti parameters, they will be used at training time only
        epsilon: 0.00001
        window: 50
        nbMaxIterations: 3000
      delft:
        # deep learning parameters
        architecture: "BidLSTM_CRF_FEATURES"
        #architecture: "scibert"
        useELMo: false
        embeddings_name: "glove-840B"
        runtime:
          # parameters used at runtime/prediction
          max_sequence_length: 3000
          batch_size: 20
        training:
          # parameters used for training
          max_sequence_length: 3000  
          batch_size: 30
```

- we only look at two system properties, optionally defined, before using the local grobid-home:
```java
private static final String PROP_GROBID_HOME = "org.grobid.home";
private static final String PROP_GROBID_CONFIG = "org.grobid.config";
```

A new documentation page "Configuration" has been added to detail the YAML config. parameters. 

Also in this PR:
- start some renaming for now inacurate terms (pdf2xml -> pdfalto, CRF -> sequence labelling)
- a bit of optimization after playing with jprofiler again, with a runtime gain around 10%

There is no more mechanism to map environment variables to config. parameters and to override them. I think it's not possible to have a general mechanism to map the rich nested parameters from env. variables, or it would be far too complicated. In particular in case of list, an environment variable will always be ambiguous without using some index or something like XPath... 
Using env. variable for configuration is only a solution for trivial environment specific configuration changes. Given that env. variables are not required for Docker, it's not blocking. 

Remaining issue: 

- for running the evaluation gradle actions, I can't get the logger using the `grobid.yaml` logger config or the `log4j.xml` in the classpath (but I've never managed to get it working before neither). All the rest (command line, server, gradle training actions) looks ok (the logger uses the unique `grobid.yaml` logger config).


de-code:
Not sure why, but GitHub seems to have issues showing the file changes for this PR. Did you rewrite the history?

kermitt2:
> Not sure why, but GitHub seems to have issues showing the file changes for this PR. Did you rewrite the history?

No, no idea why :(

de-code:
Would you be happy to create a fresh branch to help GitHub getting less confused?

kermitt2:
Yes I plan to do it next week-end.

kermitt2:
A fresh branch gives the same problem. It seems that it comes from this [commit](https://github.com/kermitt2/grobid/commit/37b0161d1913a1af49646933e451a1f268922098) where I have updated the patent processing stuff to use the parameters from the new config file and I regenerated the training data and models to have something in line with the other models... this is apparently too much for github :) 
I don't see at this point how to help GitHub.

de-code:
> A fresh branch gives the same problem. It seems that it comes from this [commit](https://github.com/kermitt2/grobid/commit/37b0161d1913a1af49646933e451a1f268922098) where I have updated the patent processing stuff to use the parameters from the new config file and I regenerated the training data and models to have something in line with the other models... this is apparently too much for github :)
> I don't see at this point how to help GitHub.

Hmm... I am a huge fan of small PRs and a convert to squash merges.
With those tainted glasses in mind, my suggestion would be to keep the whole training data and models update separate.
Even if GitHub didn't give up, it would make reviewing difficult.
In order for you to use the in-progress configuration change, you could merge the PR branch `yaml-config` into another branch with the data changes, say `patent-training-data` (and keep it up-to-date with `yaml-config` until this PR is merged into master).
What do you think?

Perhaps more specifically I would also split the patent related changes into two separate PRs, one for the code changes and another for the data and model changes (that is somewhat related to #576). That is again because a PR with changes to thousand data files would be very difficult to review.

kermitt2:
I will try to separate the modified patent data from this branch... the problem in general being often that the data/code will not work separately. 

Apparently it's actually not really related to the number of files (only around 300 files are modified), but to the size of the xml files, a patent having typically 5-10k words, so I would have expected that GitHub only doesn't display the change preview for the large files, as it does in general quite well.

lfoppiano:
Perhaps would be a good opportunity to remove all the static methods from the `GrobidProperties` class?

lfoppiano:
I'm wondering whether would be possible to provide a single string for the urls of the external services, see [here](https://github.com/kermitt2/grobid/blob/12f8c96a3d959e786b238dff836870d83ce003f3/grobid-home/config/grobid.yaml#L31). 

Instead of three properties, to provide just one? `url: https://host:port/path/path`

kermitt2:
> I'm wondering whether would be possible to provide a single string for the urls of the external services, see here.

Well I wanted to align and be consistent with Dropwizard:

https://github.com/kermitt2/grobid/blob/12f8c96a3d959e786b238dff836870d83ce003f3/grobid-home/config/grobid.yaml#L210

otherwise we're going to mix notations in the same yaml file. 

I have to say that I am using this notation a bit everywhere now for this reason :(

lfoppiano:
I see, I would say that the part you refer to the service listening ports / protocol so it makes sense to have them separated. 
In the case of URLs, I'd say the convention could be changed, and it will be much easier to manage with just one property. 
Additionally, perhaps a corner case, as it is now it won't be possible to configure a service that uses a custom port and a root path at the same time, as the output would be something like `host/root:port`

de-code:
> In the case of URLs, I'd say the convention could be changed, and it will be much easier to manage with just one property.

In case it is of any value.. I personally agree with that sentiment and prefer a single URL or path to point to external "resources".

BTW apologies, I wasn't able to review the PR yet as GitHub still seem to be lost. And the GitHub Visual Studio Code extension doesn't seem to be able to show me anything either.

kermitt2:
> In the case of URLs, I'd say the convention could be changed, and it will be much easier to manage with just one property.

Ok, I changed the notation for a simple URL. I'll try to generalize the notation elsewhere too.

kermitt2:
For configuring grobid modules: 

Let's say we want to introduce a new model in a Grobid module called `newModel`. The new model configuration can be expressed as the normal Grobid model in a yaml config file:

```
model:
  name: "newModel"
  #engine: "wapiti"
  engine: "delft"
  wapiti:
    # wapiti training parameters, they will be used at training time only
    epsilon: 0.00001
    window: 30
    nbMaxIterations: 1500
  delft:
    # deep learning parameters
    architecture: "BidLSTM_CRF"
    #architecture: "scibert"
    useELMo: false
    embeddings_name: "glove-840B"
```

In the module configuration class, we refer to the existing Grobid config class, for instance in a class `NewModuleConfiguration`:

```java
package org.grobid.core.utilities;

import org.grobid.core.utilities.GrobidConfig.ModelParameters;

public class NewModuleConfiguration {

   /* other config parameter here */ 

   public ModelParameters getModel() {
        return model;
    }

    public void getModel(ModelParameters model) {
        this.model = model;
    }
}

```

For initializing the new model, we simply do the following:

```java
        NewModuleConfiguration newModuleConfiguration = null;
        try {
            ObjectMapper mapper = new ObjectMapper(new YAMLFactory());
            newModuleConfiguration = mapper.readValue(new File("resources/config/config.yml"), NewModuleConfiguration.class);
        } catch(Exception e) {
            LOGGER.error("The config file does not appear valid, see resources/config/config.yml", e);
        }

        if (newModuleConfiguration != null && newModuleConfiguration.getModel() != null)
            GrobidProperties.getInstance().addModel(newModuleConfiguration.getModel());
        LibraryLoader.load();
```

The appropriate libraries will be loaded dynamically based on the configuration of the normal Grobid models and this new model.

de-code:
> For configuring grobid modules:

I think I don't completely understand. Will there be a separate class for every GROBID model?

But then in the following line:

```java
newModuleConfiguration = mapper.readValue(new File("resources/config/config.yml"), NewModuleConfiguration.class);
```

It appears as if `NewModuleConfiguration` is applied to the whole configuration? (now knowing how `YAMLFactory` or `ObjectMapper` works here)

kermitt2:
> I think I don't completely understand. Will there be a separate class for every GROBID model?

Mmm no it's about grobid modules such as software-mention, grobid-ner, grobid-astro, etc. Each module has its own config file (until now a property file or a json file). To make possible something consistent now in yaml (for instance now a file under `resources/config/config.yml` in the module) and given the configuration class of the module (with the configuration related to the module, in particular for one or more additional models), we can now just reuse the new grobid "core" configuration class of a model. 

For instance for software-mention: https://github.com/ourresearch/software-mentions/blob/yaml-config/resources/config/config.yml

with config class: https://github.com/ourresearch/software-mentions/blob/yaml-config/src/main/java/org/grobid/core/utilities/SoftwareConfiguration.java
(instantiated from the YAML file by the yaml parser, for example https://github.com/ourresearch/software-mentions/blob/yaml-config/src/main/java/org/grobid/service/GrobidEngineInitialiser.java#L35)
we reuse the class `ModelParameters.java`

This way, the models of the modules can be configured as the "core" grobid models, with all the range of implementation options.

de-code:
Thank you. That makes sense now.

PR #701, opened 2021-01-27
Improvement of the recovery of Pragmatic Segmenter sentence segmentation text wrt to the original text offsets

### LINKED ISSUE #753: Issue with sentence segmentation offsets
In the following example

https://arxiv.org/pdf/2103.12028v1.pdf

there are cases of wrong sentence segmentations, with sentence offsets apparently shifted by a few characters, resulting in word cut. This happens whatever the selected sentence segmenter is, OpenNLP or Pragmatic Segmenter:

```xml
<s>Human annotators evaluated the quality of document alignments for six languages (de, zh, ar, ro, et, my) selected for their different scripts and amount of retrieved documents, reporting precision of over 90%. T</s>
<s>e quality of the extracted parallel sentences is evaluated in a machine translation (MT) task on six European...</s> 
```

As it happens with both segmenters, which use different offset calculation methods, it might be due to issues with character encoding.

lfoppiano:
The pragmatic segmenter seems to modify the output string which makes the extraction of the sentence offsets more complicated. 

This PR makes two changes: 
 - Implementation using a diff-based approach to extract correctly sentence segmentation offsets when the sentences are not matching 100%
 - fix issue #753 

## Recovery of segmentation offset using Pragmatic Segmenter 
This implementation uses an external library to compute the diff of the sentence and the original text. 

For example 

``
original = "This is the    original text.   Some spaces are    going to be removed.
sentence = "This is the original text."
``

The idea is to get the starting and ending within `original` taking as reference `sentence`. 
After translating the diff in a char based string, it compute the starting and ending of the sentence within the text. 

The diff is a list of string where each element is structured as [operation, space, character], having `operation` = `+`, `-` or ` ` (e.g. comparing the string `a` and `ab` would result in the following: 
``
diff = ['  a', '-  b']
``

After the diff is computed using a two-pass heuristic:
 - starting from left, we collect all the characters starting from the first character that is common a both strings
 - starting from the right, we remove all the characters that do not equal in the diff. 

The heuristic also is limited to a subset of the string, should a sentence has been identified before.


lfoppiano:
The last commit [a577523](https://github.com/kermitt2/grobid/pull/701/commits/a5775233f00af48f8cc2ba79cccb5892d3bc699b) should fix the issue #753

Issue #697, opened 2021-01-21
Dockerfile.crf produces larger than neccessary images


superdude264:
(I've never contributed to an open-source project before, so apologies in advance.)

By making the following change to `Dockerfile.crf`, I was able to shrink the resulting container from 1.55 GB to 865 MB.
* No longer copy grobid-core-onejar.jar to the runtime stage.
* Move the unzipping of grobid-service & grobid-home as well as the zip file clean-up to the builder stage and just copying the uncompressed file.

I can submit a merge request with some or all of these changes if requested.

kermitt2:
Hi @superdude264 thanks a lot! 
This would be very useful. A pull request with the changes would be really great.

superdude264:
Awesome, I will submit early next week.

lfoppiano:
This issue seems completed 😄

PR #703, opened 2021-01-28
Dockerfile enhancements

superdude264:
https://github.com/kermitt2/grobid/issues/697

Shrunk the container from 1.66 GB to 912.5 MB. Most of the change is achieved in the first commit, which maintains total parity of output with the previous version. Later commits are remove unused artifacts, volume, & disable logging into the container (which in production environments would cause it to fail over after running for long periods).

### DIFF
--- a/Dockerfile.crf
+++ b/Dockerfile.crf
@@ -41,7 +41,6 @@ COPY grobid-trainer/ ./grobid-trainer/
 RUN ./gradlew clean assemble --no-daemon  --info --stacktrace
 
 WORKDIR /opt/grobid
-RUN cp /opt/grobid-source/grobid-core/build/libs/grobid-core-*-onejar.jar ./grobid-core-onejar.jar
 RUN unzip -o /opt/grobid-source/grobid-service/build/distributions/grobid-service-*.zip && \
     mv grobid-service* grobid-service
 RUN unzip -o /opt/grobid-source/grobid-home/build/distributions/grobid-home-*.zip && \
--- a/Dockerfile.crf
+++ b/Dockerfile.crf
@@ -22,9 +22,6 @@ RUN apt-get update && \
 
 WORKDIR /opt/grobid-source
 
-RUN mkdir -p .gradle
-VOLUME /opt/grobid-source/.gradle
-
 # gradle
 COPY gradle/ ./gradle/
 COPY gradlew ./
@@ -59,12 +56,6 @@ RUN apt-get update && \
 WORKDIR /opt/grobid
 
 COPY --from=builder /opt/grobid .
-
-# below to allow logs to be written in the container
-# RUN mkdir -p logs
-
-VOLUME ["/opt/grobid/grobid-home/tmp"]
-RUN chmod 777 /opt/grobid/grobid-home/tmp
 
 ENV JAVA_OPTS=-Xmx4g
 
--- a/Dockerfile.crf
+++ b/Dockerfile.crf
@@ -47,7 +47,7 @@ RUN rm -rf grobid-source
 # -------------------
 # build runtime image
 # -------------------
-FROM openjdk:8u212-jre-slim
+FROM openjdk:11-jre-slim
 
 RUN apt-get update && \
     apt-get -y --no-install-recommends install libxml2 && \
@@ -56,8 +56,6 @@ RUN apt-get update && \
 WORKDIR /opt/grobid
 
 COPY --from=builder /opt/grobid .
-
-ENV JAVA_OPTS=-Xmx4g
 
 # Add Tini
 ENV TINI_VERSION v0.18.0
--- a/Dockerfile.crf
+++ b/Dockerfile.crf
@@ -18,7 +18,7 @@ FROM openjdk:8u212-jdk as builder
 USER root
 
 RUN apt-get update && \
-    apt-get -y --no-install-recommends install libxml2
+    apt-get -y --no-install-recommends install unzip
 
 WORKDIR /opt/grobid-source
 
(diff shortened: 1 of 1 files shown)


grobid-service/config/config.yaml
```diff
@@ -30,11 +30,4 @@ logging:
   appenders:
     - type: console
       threshold: ALL
-      timeZone: UTC
```
lfoppiano, line 33: This change would impact the log file in general, I would make the modification on the fly when creating the image

lfoppiano (changes requested):
Looks good, although I would not change the config file

grobid-service/config/config.yaml
superdude264: The intention was not log to a file at all as that will eventually fill up the allocated storage space for the container. If the preference is to have a log file by default, I can drop this change. Making the change at container build time in a robust manner would require adding a YAML parser to the builder (e.g. https://github.com/biojppm/rapidyaml).  Is that OK?

superdude264: @lfoppiano Any thoughts?

kermitt2: Hello ! 

In the next GROBID version, this service config file will disappear and will be replaced by a unique yaml config file for Grobid (replacing the grobid.properties too), see PR #687 

Then we can modify and mount the config file at launch of the docker container:
> docker run --rm --gpus all --init -v /home/lopez/grobid/grobid-home/config/grobid.yaml:/opt/grobid/grobid-home/config/grobid.yaml:ro  grobid/grobid:0.6.2-SNAPSHOT

and this way the log will be defined for the service at the container run time, which is more flexible.  

So I would say the default log is less important (personally I don't like the logs on the console, which I find unusable).

superdude264: Those new config changes are great. In our production setup we modify both config.yaml & grobid.properties before copying them in to the container, so having just one file to manage will be an improvement.

I'll drop my change to config.yaml here since they are somewhat orthogonal to the size issue.

lfoppiano: Just for references, my idea was to keep the original config and adapt it based on the needs in docker. For example removing that lines in the config file via bash scripting when building the docker image.  

Anyway, indeed with the config file passed to the container the problem is solved at the source ;-)

superdude264:
Please let me know if there are any other changes you guys would like me to make to this MR.

lfoppiano (approved):
looks good to me

superdude264:
Should I upgrade to the base images to Java 11?  I was able to do that locally with no issues.  Java version >11 had issues though.

superdude264:
Additionally, tini is no longer required for Docker 1.13+: https://github.com/krallin/tini

You can see in the screenshot below that docker-init (built-in tini) is forked to tini.

image

@lfoppiano @kermitt2 Do you guys want me to continue pushing updates here?

lfoppiano:
@superdude264 ok for me ✅

superdude264:
@lfoppiano The only other change I have to upgrade the runtime image to Java 11.

lfoppiano:
@superdude264 you can include the update to java 11 in the current PR I think

superdude264:
Here you go @lfoppiano. I left the builder image as-is and upgraded the runtime image to Java 11. The JVM now utilizes the memory assigned to the container, so there isn't a need to set the memory size in `$JAVA_OPTS`.  There are some new informational warnings on startup:
```
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by com.google.inject.internal.cglib.core.$ReflectUtils$1 (file:/opt/grobid/grobid-service/lib/guice-4.1.0.jar) to method java.lang.ClassLoader.defineClass(java.lang.String,byte[],int,int,java.security.ProtectionDomain)
WARNING: Please consider reporting this to the maintainers of com.google.inject.internal.cglib.core.$ReflectUtils$1
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release
```
I have not found a way to correct or silence these warning in our production setup, so I just left them.

Final image size is ~950 MB.

kermitt2:
Thank you @superdude264 (and Luca!), it works fine for me too.

I started to adapt the changes to `Dockerfile.delft` to have something consistent (the DL dockerfile is more complicated and leads to a considerably larger image), which I will commit separately. If you can have a look to it too maybe, it would be great, otherwise no problem - it will be a very fat image anyway ;)

kermitt2:
@superdude264 I have a question, you removed the volume for the GROBID `tmp` space:

```
VOLUME ["/opt/grobid/grobid-home/tmp"]
```  

I've read somewhere that writing in a volume is faster than in the container directly, but maybe this is not the case. Do you have any hints on this?

superdude264:
@kermitt2 Re: `Dockerfile.delf`, point me to the branch or tag me in an MR and I will take a look.

I will look into the question regarding write speed to a container vs a volume.

kermitt2:
Thanks! see https://github.com/kermitt2/grobid/pull/721

superdude264:
### RE: Removal of `VOLUME ["/opt/grobid/grobid-home/tmp"]`

The [`VOLUME` command]( https://docs.docker.com/engine/reference/builder/#volume) creates a volume on the host (most of the time; see caveats below) and the Docker documentation advocates using volumes for write-heavy paths [here](https://docs.docker.com/storage/storagedriver/#copying-makes-containers-efficient) & [here](https://docs.docker.com/storage/storagedriver/overlayfs-driver/#performance-best-practices).

However, I would recommend using a [tmpfs mount](https://docs.docker.com/storage/tmpfs/) by adding ` --tmpfs /opt/grobid/grobid-home/tmp` to the `docker run` commands in the[ GROBID with containers documentation](https://grobid.readthedocs.io/en/latest/Grobid-docker/). This will back that directory using [tmpfs](https://www.wikiwand.com/en/Tmpfs) and result in better performance than writing to disk. 

Additionally, it may be worthwhile to have grobid default to writing to the system temporary directory (using `Files.createTempDirectory()`), as this will normally be setup w/ tmpfs (in most Linux distributions running on bare metal) or container operators will create a tmpfs mount there as a best practice.

#### Caveats
I have not tested the above because I run grobid in AWS Fargate, which only supports network-mounted volumes (AWS Elastic File System) and so ignores `VOLUME` directives in Dockerfiles. Additionally, I do my development on macOS and [tmpfs is Linux-only](https://docs.docker.com/storage/tmpfs/#limitations-of-tmpfs-mounts).

PR #707, opened 2021-02-05
Add more tests on sentence segmentation

lfoppiano:
I'm adding more tests for testing that the sentence segmentation correction works fine. 

Actually I though there was a bug, but then after I've added tests I realised it was fine. The tests have more realistic examples and they allow to be independent than the sentence segmenter (which might change the results in future).

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -203,7 +203,7 @@ subprojects {
     test {
         testLogging.showStandardStreams = true
         // enable for having separate test executor for different tests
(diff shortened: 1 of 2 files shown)


kermitt2:
Thanks Luca! Very good I think that it's testing the functionality of "forbidden" segmentation places, and not the quality of the segmenter (then it would need rather some benchmarking).

kermitt2:
However the tests are failing :/

PR #714, opened 2021-02-15
fixed fulltext block start

### LINKED ISSUE #712: Full text model layout features: BLOCKSTART missing, if very first block token is a new line
At least for some documents, the first token of a block seem to be a line feed.

In that case the line feed is [filtered out](https://github.com/kermitt2/grobid/blob/0.6.1/grobid-core/src/main/java/org/grobid/core/engines/FullTextParser.java#L695-L698):

```java
	                if (TextUtilities.filterLine(text)) {
						n++;
	                    continue;
	                }
```

But when it is then getting to process the next "real" token, `n` will no longer be `0` but `1`. Therefore it will not go into the [main blockstart block](https://github.com/kermitt2/grobid/blob/0.6.1/grobid-core/src/main/java/org/grobid/core/engines/FullTextParser.java#L768-L778):

```java
	                if (n == 0) {
	                    features.lineStatus = "LINESTART";
	                    // be sure that previous token is closing a line, except if it's a starting line
	                    if (previousFeatures != null) {
	                    	if (!previousFeatures.lineStatus.equals("LINESTART"))
		                    	previousFeatures.lineStatus = "LINEEND";
	                    }
	                    if (token != null)
		                    lineStartX = token.getX();
	                    features.blockStatus = "BLOCKSTART";
	                } else if (n == tokens.size() - 1) {
```

---

## Example document

`475335v1` (DOI: [10.1101/475335](https://doi.org/10.1101/475335))

<details>
<summary>PDF</summary>

image
</details>

<details>
<summary>bioRxiv XML</summary>

```xml
<sec id="s3c">
    <title>Epidemic synchrony and annual phase coherence</title>
    <p>
        We explored correlations between dengue time series in different regions. Both epidemic synchrony and phase coherence were higher for closer regions and declined with distance (
        <xref rid="fig3" ref-type="fig">Fig 3</xref>
        ). For the Urban-2 (n = 161) spatial level, epidemic synchrony reached the average countrywide correlation at approximately 1,260 kilometres (
        <xref rid="fig3" ref-type="fig">Fig 3A</xref>
        ). This synchrony length represents a substantial part of Brazil’s dimensions as the country extends 4,395 kilometres north to south and 4,319 kilometres west to east. The coherence length had a higher value of 1,590 kilometres (
        <xref rid="fig3" ref-type="fig">Fig 3B</xref>
        ), suggesting that agreement in dengue seasonality spreads further than correlations of epidemic curves.
    </p>
    <fig id="fig3" position="float" fig-type="figure">
        <label>Fig 3.</label>
        <caption>
            <title>
                Epidemic synchrony and annual phase coherence between Brazilian Urban-2 regions.
            </title>
            <p>
                Epidemic synchrony (A) and annual phase coherence (B) summarised using nonparametric spline covariance function. Solid blue line describes the mean pairwise correlation from the data and the dotted lines represent the 95% envelope for bootstrapped correlations of case and annual phase angle time series, respectively. Red line indicates global countrywide correlation.
            </p>
        </caption>
        <graphic xlink:href="475335_fig3.tif"/>
    </fig>
    <p>
        We also looked at epidemic synchrony and annual phase coherence at other spatial levels (S4 and S5 Figs, respectively) and found that both synchrony and coherence lengths tend to decrease for smaller spatial resolutions and stabilise at 1,240 km and 1,500 km.
    </p>
</sec>
```
</details>

The text `We also looked at epidemic synchrony...` (line 216) doesn't get the `BLOCKSTART` feature (it will be `BLOCKIN`), even though it is in its own block (but with a line feed as the first token as described above).

---

I could try to submit a fix PR for it.

/cc @kermitt2

de-code:
resolves #712

### DIFF
--- a/grobid-core/src/test/java/org/grobid/core/engines/FullTextParserTest.java
+++ b/grobid-core/src/test/java/org/grobid/core/engines/FullTextParserTest.java
@@ -4,6 +4,7 @@ import org.apache.commons.lang3.tuple.Pair;
 import org.grobid.core.analyzers.GrobidAnalyzer;
 import org.grobid.core.document.Document;
 import org.grobid.core.document.DocumentPiece;
+import org.grobid.core.document.DocumentPointer;
 import org.grobid.core.factory.GrobidFactory;
 import org.grobid.core.layout.Block;
 import org.grobid.core.layout.LayoutToken;
@@ -13,12 +14,17 @@ import org.grobid.core.utilities.GrobidProperties;
 import org.junit.AfterClass;
 import org.junit.Before;
 import org.junit.BeforeClass;
+import org.junit.Ignore;
 import org.junit.Test;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.List;
 import java.util.SortedSet;
+import java.util.TreeSet;
 import java.util.stream.Collectors;
 
 import static org.easymock.EasyMock.*;
@@ -28,6 +34,8 @@ import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.collection.IsCollectionWithSize.hasSize;
 
 public class FullTextParserTest {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(FullTextParserTest.class);
 
     private FullTextParser target;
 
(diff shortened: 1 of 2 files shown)


de-code:
I am currently running into the issue described in https://github.com/kermitt2/grobid/issues/712#issuecomment-779275621
I guess the way I am creating the test data isn't quite correct (using `createFromText` and `getDocumentPartsForLayoutTokens`, the latter extracted from `processShortNew`).

de-code:
> I am currently running into the issue described in [#712 (comment)](https://github.com/kermitt2/grobid/issues/712#issuecomment-779275621)
> I guess the way I am creating the test data isn't quite correct (using `createFromText` and `getDocumentPartsForLayoutTokens`, the latter extracted from `processShortNew`).

I got around the problem by manually creating `DocumentPiece` for the whole document. I therefore reverted `getDocumentPartsForLayoutTokens` as it wouldn't be used by the test and could be refactored separately.

grobid-core/src/test/java/org/grobid/core/engines/FullTextParserTest.java
```diff
@@ -1,30 +1,42 @@
 package org.grobid.core.engines;
 
+import org.apache.commons.lang3.tuple.Pair;
 import org.grobid.core.analyzers.GrobidAnalyzer;
```
de-code, line 4: There are quite a few unused imports. Might be good to tidy it up (and remove commented out code, perhaps mark test to be ignored instead).

grobid-core/src/test/java/org/grobid/core/engines/FullTextParserTest.java
```diff
@@ -43,6 +55,41 @@ public static void tearDown() {
         GrobidFactory.reset();
     }
 
+    public DocumentPiece getWholeDocumentPiece(Document doc) {
```
de-code, line 58: maybe `getWholeDocumentPiece` and `getWholeDocumentParts` could be moved to a central place, if it doesn't exist already.

de-code:
Other parsers are potentially affected as well.

There appear to be a lot of duplication and and could probably be refactored.

de-code:
Raised suggestion for refactoring the features: https://github.com/kermitt2/grobid/issues/718

lfoppiano (commented):
I'm having trouble understanding the whole picture here, so I'm not sure I can really review this part.

de-code:
> I'm having trouble understanding the whole picture here, so I'm not sure I can really review this part.

There is an example in the issue #712 - happy to add more information?

kermitt2:
I have regenerated the fulltext model feature files with the fix, but actually there is no difference with before with respect to starting block. I also retrained the model (before checking the difference in features), in branch https://github.com/kermitt2/grobid/tree/fix-712
So apparently the problem in #712 never applies in practice in the training data at least. 

It might appear in some new files like in your example, but the benchmark on PMC set only changes at the second decimal, so it should be minor. 

So we could simply merge the fix, but no need to update the training files and model for this.

PR #721, opened 2021-02-22
Aligned Deep Learning Dockerfile with PR #703

kermitt2:
I tried to adapt the change to the CRF-only dockerfile to the one using DeLFT, see PR #703 . This is reducing the image in a similar manner from 10.3GB to 9.8GB, but of course this is less impressive at this scale :)

Still using JRE 8 in the current form.

### DIFF
--- a/Dockerfile.delft
+++ b/Dockerfile.delft
@@ -36,6 +36,15 @@ COPY settings.gradle ./
 COPY grobid-home/ ./grobid-home/
 COPY grobid-core/ ./grobid-core/
 COPY grobid-service/ ./grobid-service/
+COPY grobid-trainer/ ./grobid-trainer/
+
+# cleaning unused native libraries before packaging
+RUN rm -rf grobid-home/pdf2xml/lin-32
+RUN rm -rf grobid-home/pdf2xml/mac-64
+RUN rm -rf grobid-home/pdf2xml/win-*
+RUN rm -rf grobid-home/lib/lin-32
+RUN rm -rf grobid-home/lib/win-*
+RUN rm -rf grobid-home/lib/mac-64
 
 RUN ./gradlew clean assemble --no-daemon  --info --stacktrace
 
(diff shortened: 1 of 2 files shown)


Dockerfile.delft
```diff
@@ -60,33 +61,25 @@ ENV LANG C.UTF-8
...
+COPY --from=builder /opt/grobid .
 
-RUN unzip ./grobid-home.zip -d ./grobid && \
-    mkdir -p /opt/grobid/grobid-home/tmp
-RUN rm *.zip
 RUN rm -rf /opt/grobid/grobid-home/pdf2xml/lin-32
 RUN rm -rf /opt/grobid/grobid-home/pdf2xml/mac-64
 RUN rm -rf /opt/grobid/grobid-home/pdf2xml/win-*
 RUN rm -rf /opt/grobid/grobid-home/lib/lin-32
 RUN rm -rf /opt/grobid/grobid-home/lib/win-*
```
superdude264, line 77: I would move this up into the builder.

In Docker, each RUN, ADD, or COPY command create a tarball behind the scenes and all of these are bundled together to form a stage (e.g builder or runner). The final stage is what determines the size of the resulting container. Thus it's better to remove files in the builder, then copy to the runner.

Dockerfile.delft
```diff
@@ -60,33 +61,25 @@ ENV LANG C.UTF-8
...
 
-# below to allow logs to be written in the container
+# below to allow logs to be written in the container (not advised for production!)
 # RUN mkdir -p logs
```
superdude264, line 80: I've found that even if this directory is not present, log files will still be written to the container. To not write the logs `/opt/grobid/grobid-service/config/config.yaml` has to be updated so that the only appender is `CONSOLE`.  This was discussed in #703, but that change was not made.

Dockerfile.delft
```diff
@@ -60,33 +61,25 @@ ENV LANG C.UTF-8
...
 
+# the following might not be necessary - but should result in faster temp file write than directly in the container
 VOLUME ["/opt/grobid/grobid-home/tmp"]
```
superdude264, line 83: See https://github.com/kermitt2/grobid/pull/703#issuecomment-783388362

Dockerfile.delft
kermitt2: what a good advice, this is brilliant ! I decreased the size of the crf image to 650MB doing that.

Dockerfile.delft
kermitt2: Thank you ! I am removing the VOLUME and opening an issue for using `Files.createTempDirectory()`.
The GROBID tmp is not really read/write intensive - just write/read/delete the PDF and the ALTO file, which are negligible as compared to the whole process.

kermitt2:
Reduction to 9.56GB so far... I think we can decrease by a bit more than 2GB by managing better the embeddings, but I leave that for a future release!

Issue #705, opened 2021-02-01
No retry mechanism on CrossrefRequest results in hanging.


AaronNGray:
There does not seem to be any retry mechanism in CrossrefRequest resulting in it hanging overall Grobid REST requests.

lfoppiano:
@AaronNGray The PR #725 introduces various improvements to the mechanism for calling Crossref, including adding the timeout. Feel free to have a look at it if you like.

elonzh:
I am using v0.6.2 and sometimes CrossrefRequest will be blocked.

The grobid will hang and can not release resources.

![image](https://user-images.githubusercontent.com/8749411/124360745-f0813180-dc5d-11eb-8507-48cc79f3208f.png)

![image](https://user-images.githubusercontent.com/8749411/124360730-d34c6300-dc5d-11eb-8438-71b0d3a03843.png)

![image](https://user-images.githubusercontent.com/8749411/124360737-dba49e00-dc5d-11eb-8570-249e0859ac27.png)

kermitt2:
Normally it's fixed with #725, which is addressing this. 
(I don't know why the milestone has been changed to 0.6.2, but it's in 0.7.0 release - or master - as indicated initially)

kermitt2:
>  (I don't know why the milestone has been changed to 0.6.2, but it's in 0.7.0 release - or master - as indicated initially)

double check and it's in 0.6.2 indeed, so the timeout should apply and release the thread, it's weird.

elonzh:
It seems a quite serious problem, consolidation with crossref almost always makes grobid hang and can't return response.

kermitt2:
mmm I just process 2000 pdf with crossref consolidation for header without issue, just a bit slower than biblio-glutton. Independently from the timeout which should apply and apparently it is not, do you indicate a `mailto` in `config/grobid.yml` for the polite use of CrossRef ?

elonzh:
> mmm I just process 2000 pdf with crossref consolidation for header without issue, just a bit slower than biblio-glutton. Independently from the timeout which should apply and apparently it is not, do you indicate a `mailto` in `config/grobid.yml` for the polite use of CrossRef ?

mmmm, try to consolidate citations.

elonzh:
I will try to reproduce it tomorrow.

kermitt2:
Consolidating citations now, with 8 threads, (crossref `mailto` address in the config file), it becomes slow after a while, Crossref service lowers the availability, but still going on. 

```
[...]

PR #725, opened 2021-03-03
Updates to the CrossRef requests implementation 

Aazhar:


lfoppiano:
@Aazhar is this PR related to crossref? Could you check this issue too? #705

kermitt2:
Adding some background about the problem: 

The CrossRef support desk has indicated that the max query rate to follow by the CrossRef client, which is given in the header pf the crossref response is not reliable - at least not reliable anymore, e.g. 50 requests max per second. So we can't use anymore the existing TimedSemaphore implementation which dynamically adapt the query rate of the CrossRef client to the indicated max query rate based on time range.  

They indicated that we need rely only on the total number of active parallel queries that should not go beyond the limit given in the indicated `X-Rate-Limit-Limit`. So the PR uses now a non-bounded `CachedThreadPool` and we check that the limit of allowed active threads/queries is lower than this limit. 

@Aazhar you mentioned `X-LIMIT-LIMIT`, but it's actually still `X-Rate-Limit-Limit` right? We don't capture a `X-LIMIT-LIMIT` in the response currently.

lfoppiano:
Thanks, @kermitt2. Meanwhile, I set this PR as "draft" as it seems not yet ready to be merged

kermitt2:
> Could you check this issue too? #705

#705 is related, yes - this PR fixes a serious issue when scaling, leading possibly to an ever-growing of threads. But with the fixed number of max queries that is introduced now, having ever hanging queries without CrossRef response might fill all the allowed number of threads - so #705 is important to fix too. 

For issue #705 I think we need to check the futures to see if they finished normally or canceled by timeout and re-send the query, it's a bit complicated. I thought there was some default timeout with `CachedThreadPool` but apparently not. 

Of course a good solution is to use biblio-glutton rather than CrossRef :D

kermitt2:
I removed the division of the rate by 2, its not relevant anymore because we are not using the sliding time window for the number of active threads. Keeping it cause the number of threads to be always half of the CrossRef response indicated `X-Rate-Limit-Limit` and processing is simply 2 times longer...

Tested with the 1942 PDF files of the PMC sample with full reference consolidation, runtime is now the same as with the previous `TimedSemaphore` implementation (between 150 and 180 minutes).

Aazhar:
hello @lfoppiano 
regarding #705 , grobid is using a cached thread pool in order to parallelize crossref request, the static Executors.newCachedThreadPool is implemented in the standard Java library:
```java
public static ExecutorService newCachedThreadPool() {
    return new ThreadPoolExecutor(0, Integer.MAX_VALUE,
                                  60L, TimeUnit.SECONDS,
                                  new SynchronousQueue<Runnable>());
}
```

there is a timeout of  60 seconds , so no request should be hanging

Aazhar:
> Tested with the 1942 PDF files of the PMC sample with full reference consolidation, runtime is now the same as with the previous `TimedSemaphore` implementation (between 150 and 180 minutes).

do you know in average how long is the duration of a request thread  for you ?

kermitt2:
I changed the sleep time for a new crossref request thread when the pool is full to `TimeUnit.MICROSECONDS.sleep(10);` after some tests... not super sure because crossref service response time is changing over the day apparently (as my home internet bandwidth), but it seems to improve a bit (by 10-15 minutes for the 90K references).   

The timeout and failure aspects should really be improved, with logging of crossref request timeout, retry, etc.... Currently we are not even sure if all the crossref queries are timing out or really answered as expected (when looking at the TEI files,we see that it is working)... it's overall really very slow when consolidating all references, in particular as compared to biblio-glutton.

PR #726, opened 2021-03-05
Minor correction on citation features

### LINKED ISSUE #720: Minor: citation type
I bumped into this minor thing by chance (see [here](https://github.com/kermitt2/grobid/blob/bfc10f77998a141e14e1a88483ef42954e0d30b2/grobid-core/src/main/java/org/grobid/core/engines/FullTextParser.java#L502), shouldn't be `nbAuthorType` instead of `nbNumbType`?

lfoppiano:
Resolves #720


kermitt2:
The commit for changing to `nbAuthorType` is missing in this PR... it only contains the retrained model no?

kermitt2:
... so I have to retrain and re-test the `fulltext` model with the missing change :) 
(it seems then that it explains well why it makes no difference!)

lfoppiano:
ahhhh.... sorry... I might have forgotten to push? :facepalm:

I will check

kermitt2:
No it's good, it's just that we don't see the change for the `FullTextParser,java` in the "Files changed" of the PR, but looking at the file of this branch the modification is present (confusing I think). 

So the retrained model is correct!

kermitt2:
And finally the change is not present after the merge with master... really confusing...

PR #727, opened 2021-03-08
Remove the logging while building the docker image

lfoppiano:
This PR removes the log on a file of the application. It's performed at runtime while building so that the log file can be kept for other usages. 

It requires just that the two marks in the `config.yml` are not removed or modified

### DIFF
--- a/Dockerfile.crf
+++ b/Dockerfile.crf
@@ -4,7 +4,8 @@
 
 ## See https://grobid.readthedocs.io/en/latest/Grobid-docker/
 
-## docker build -t grobid/grobid:GROBID_VERSION --build-arg GROBID_VERSION=GROBID_VERSION .
+## docker build -t grobid/grobid:GROBID_VERSION --build-arg GROBID_VERSION=GROBID_VERSION -f Dockerfile.crf .
+## docker build -t grobid/grobid:GROBID_VERSION --build-arg GROBID_VERSION=GROBID_VERSION -f Dockerfile.delft .
 ## docker run -t --rm -p 8080:8070 -p 8081:8071 {image_name}
 
 # To connect to the container with a bash shell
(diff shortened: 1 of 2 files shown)


superdude264:
@lfoppiano Looks good. Will definitely help users avoid some surprises when they use the image from DockerHub.

kermitt2:
I am not sure it is a good idea:
- we should reasonably expect logging when using the docker image, because it's typically a production usage
- I think it's confusing for the user to have logging settings specified for the service and then not using it
- all the config will change in a unique yaml file with #687 and I plan to merge to this version (which is changing a lot of things) immediately after releasing 0.6.2, so this removal of the log will have to be changed

If we want to avoid logs in the container, why don't we simply use a volume which will be managed on the host:

```
VOLUME ["/opt/grobid/logs"]
```

Nothing else to do I think. This is also a good choice for persistent data and it easier than bind mount which requires something in the command line.

This would allow logging and avoid automatically modifying the config files, which is a bit confusing for the user who sets the settings.

superdude264:
The abbreviated idea around Docker logging is that the Docker runtime captures the logs using a [log driver](https://docs.docker.com/config/containers/logging/configure/). In the simplest (and I believe default) case, Docker captures standard out & standard error and stores them, so the console logs will still be available.

Additionally, putting the `VOLUME` keyword in the Dockerfile is not a guarantee that it will be created by the Docker runtime on the host. That decision depends on how the operator sets it up. If they don't create a volume and mount it, then the container will just write the log data into itself and it will be removed when the container is terminated. In a production setting, this would likely mean the runtime would kill the container for taking up too much space (because of all the log data). This is what happened in our production setup.

lfoppiano:
This is not anymore valid, since we don't have log files in docker by default.

PR #715, opened 2021-02-16
Fix NullPointer exception in PDFALTOOutlineSaxHandler

lucbouge:
I observed the following error. 

févr. 12, 2021 3:20:44 PM org.grobid.core.document.Document addTokenizedDocument
GRAVE: Cannot parse file: my_dir/grobid-0.6.1/grobid-home/tmp/ZW3PpM4E9g.lxml_outline.xml

févr. 12, 2021 3:20:44 PM org.grobid.core.document.Document addTokenizedDocument
GRAVE: Cannot parse file: my_dir/grobid-0.6.1/grobid-home/tmp/ZW3PpM4E9g.lxml_outline.xml

févr. 12, 2021 3:20:46 PM org.grobid.core.engines.ProcessEngine createTraining
INFOS: 2 files processed.

it originates in an illegal nullPointer access in file PDFALTOOutlineSaxHandler.java. 

In understand that the error is not  that "GRAVE". According to @kermitt2, it is simply one of the generated XML file resulting from the PDF parsing which is not XML valid (very frequent) - it has no impact because the outline file (containing the table of content outline if available embedded in the PDF) is not exploited by GROBID for the moment - it's to allow some possible improvements in the future. The error is more a reminder for the developers... the XML parser that classifies it as "GRAVE" but it would be rather INFO for us.

Unfortunately, a side effect might have been overlooked. If father is null, then father.addChild(currentNode) is called with a nullPointer exception.

This exception is caught by catch (Exception e) at line 372 in grobid-core/src/main/java/org/grobid/core/document/Document.java where the error message "Cannot parse" is misleading.

I think an additional else is just missing.

PR #729, opened 2021-03-09
 Fix NullPointer exception in PDFALTOOutlineSaxHandler

lfoppiano:
... update of #715 

CC @lucbouge


grobid-core/src/main/java/org/grobid/core/sax/PDFALTOOutlineSaxHandler.java
```diff
@@ -32,6 +32,7 @@
 	private int currentParentId = -1;
 
 	private Map<Integer,DocumentNode> nodes = null;
+	private Map<Integer, Integer> labels = new HashMap<>();
```
kermitt2, line 35: This seems not used?

grobid-core/src/main/java/org/grobid/core/sax/PDFALTOOutlineSaxHandler.java
```diff
@@ -53,11 +54,14 @@ public void endElement(java.lang.String uri, java.lang.String localName,
 			java.lang.String qName) throws SAXException {
 
 		if (qName.equals("STRING")) {
-			label = getText();
+		    currentNode.setLabel(getText());
+            currentNode.setBoundingBox(box);
```
kermitt2, line 58: bounding box is not yet defines at the level of STRING, it's when ITEM is closed that we know for sure (see previous version)

grobid-core/src/main/java/org/grobid/core/sax/PDFALTOOutlineSaxHandler.java
```diff
@@ -53,11 +54,14 @@ public void endElement(java.lang.String uri, java.lang.String localName,
...
 		} else if (qName.equals("ITEM")) {
-			currentNode.setLabel(label);
-			currentNode.setBoundingBox(box);
-		}
+            box = null;
+            label = null;
+		} else if (qName.equals("TOCITEMLIST")) {
+		    currentParentId = -1;
```
kermitt2, line 63: This is actually  50% of the origin of the error...  currentParentId was not reinit.

grobid-core/src/main/java/org/grobid/core/sax/PDFALTOOutlineSaxHandler.java
```diff
@@ -78,7 +82,7 @@ public void startElement(String namespaceURI, String localName,
 				String value = atts.getValue(i);
 
 				if ((name != null) && (value != null)) {
-					if (name.equals("id")) {
+					if (name.equalsIgnoreCase("id")) {
```
kermitt2, line 85: yes :) 
a bug - that was leading to errors

grobid-core/src/main/java/org/grobid/core/document/DocumentNode.java
```diff
@@ -182,5 +184,12 @@ else if (father == null) {
              }
          }
      }*/
+    public Integer getId() {
+        return id;
+    }
+
+    public void setId(Integer id) {
+        this.id = id;
+    }
```
kermitt2, line 193: not used?

grobid-core/src/main/java/org/grobid/core/sax/PDFALTOOutlineSaxHandler.java
lfoppiano: Yes, it's not used...

grobid-core/src/main/java/org/grobid/core/document/DocumentNode.java
lfoppiano: it's used at line 94 of the `PDFALTOOutlineSaxHandler.java `

grobid-core/src/main/java/org/grobid/core/sax/PDFALTOOutlineSaxHandler.java
lfoppiano: good point. I update this and include the bounding box in the text. However, if you check the test `testParsing_pdf2XMLOutline_errorcase_ShouldWork` the bounding box values (e.g. left and right seems to be inverted?)

kermitt2: I didn't change the left and right attributes from pdf2xml, but at the time I was not 100% sure how to set them. I looked at some examples and right now it's:

```
                        double x = left;
			double y = right;
			double width = -1.0;
			double height = -1.0;
			if (right >= left)
				width = right - left;
			if (bottom >= top)
				height = bottom - top;
			box = BoundingBox
				.fromPointAndDimensions(page, x, y, width, height);
```

From your example, it's not working and it would be rather:  `top` is `y`, `bottom` is `height`, `left` is `x`, right is `width`. 
I am not sure it's consistent from one PDF to another one. Ideally it should be corrected in pdfalto (there's actually some piece of conversion into x/y in pdfalto for this, but it's finally not outputted :) ).

lfoppiano: Great! Thanks! I've updated the code to follow the new convention. 
I've also fixed another bug occurring when adding the bounding box at the end of the `<ITEM>` will not work for nested structure, I now add the bounding box right at the end of `<LINK>`

BTW: In the example also the pages don't seem consistent (lower items point to the first page), but it's a bug in the PDF.

kermitt2: OK sorry I was not clear enough, the correspondence I indicated was only for your new example, it's going to clash with others. The way it was set first works for some documents and should be taken into account. In general I think it should be addressed in pdfalto considering all the different cases we can observe in a PDF.

kermitt2: > BTW: In the example also the pages don't seem consistent (lower items point to the first page), but it's a bug in the PDF.

Yes this outline example has actually multiple bugs/problems, it should probably not be considered for general test :D

kermitt2: > I've also fixed another bug occurring when adding the bounding box at the end of the `<ITEM>` will not work for nested structure, I now add the bounding box right at the end of `<LINK>`

Good point yes !

lfoppiano: OK. I reverted the change on the bounding box. 
I also remove the bounding box values from the test, as they don't make much sense at the moment.. 
Even the other test case for the outline that I have `pdfalto.xml_outline.xml` doesn't seem totally correct. 
Perhaps I will add an issue on PDFalto about it.

grobid-core/src/test/java/org/grobid/core/sax/PDFALTOOutlineSaxHandlerTest.java
```diff
@@ -0,0 +1,74 @@
...
+
+        mockDocumentSource = createMock(DocumentSource.class);
+
+        document = Document.createFromText("");
+        target = new PDFALTOOutlineSaxHandler(document);
+    }
+
+    @Test
+    public void testParsing_pdf2XMLOutline_ShouldWork() throws Exception {
+        InputStream is = this.getClass().getResourceAsStream("pdfalto.xml_outline.xml");
```
abidaJustice, line 39: I detect that this code is problematic. According to the [Bad practice (BAD_PRACTICE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#bad-practice-bad-practice), [UI: Usage of GetResource may be unsafe if class is extended (UI_INHERITANCE_UNSAFE_GETRESOURCE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#ui-usage-of-getresource-may-be-unsafe-if-class-is-extended-ui-inheritance-unsafe-getresource).
Calling this.getClass().getResource(...) could give results other than expected if this class is extended by a class in another package.

lfoppiano: Hi, thanks for your feedback. This is the code of a unit test, it's running only after compilation and it's not supposed to be extended. Do you have any suggestion on how to change it?

abidaJustice: Since this class is not extended by a class, it should be safe. Thanks a lot!

abidaJustice: thank you for your quick reply. Since this is a test class, this problem is inapplicable

kermitt2:
Ready to merge, no?

lfoppiano:
Yes 👍

PR #738, opened 2021-04-14
added annex figures, tables, equations

### LINKED ISSUE #737: Figures and tables in the back / annex section ignored
This is related to #698

Some documents have main figures and supplementary figures.
If in those cases, the segmentation model labels the supplementary figures as `annex`,
then the content is passed separately to the `fulltext` model.
If the `fulltext` then correctly labels it as `figure`, then the figures from the `annex` are not included in the output.

de-code:
resolves #737

### DIFF
--- a/grobid-core/src/main/java/org/grobid/core/document/TEIFormatter.java
+++ b/grobid-core/src/main/java/org/grobid/core/document/TEIFormatter.java
@@ -1131,6 +1131,9 @@ public class TEIFormatter {
                                     BiblioItem biblio,
                                     List<BibDataSet> bds,
                                     List<LayoutToken> tokenizations,
+                                    List<Figure> figures,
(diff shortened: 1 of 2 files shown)


de-code:
I am not quite sure how I managed to increase the coverage without adding tests or deleting code.

lfoppiano:
With the standard model, I cannot see any difference, the "Extended Data Figure 1" is not extracted at all in both cases. 
I've checked the code and the PR looks good to me.

lfoppiano (approved):

de-code:
I noticed one issue with it, is that the figure ids are reset. e.g. first body figure has the id `fig_0`. Then the first back section / annex figure also has the id `fig_0`.

lfoppiano:
I close this, which is continued on #1215

PR #750, opened 2021-04-23
improving methods that clean doi for consolidation

lfoppiano:
As mentioned in https://github.com/kermitt2/biblio-glutton/issues/55 

When I wrote the test I realised that the result of `substring()`, aiming at removing the prefix, was not used. 

This should: 
 - remove diactric (not sure is needed)
 - remove characther that combined with others 
 - fix substring(4) using the 

About diactricts: https://memorynotfound.com/remove-accents-diacritics-from-string/

### DIFF
--- a/grobid-core/src/test/java/org/grobid/core/data/BiblioItemTest.java
+++ b/grobid-core/src/test/java/org/grobid/core/data/BiblioItemTest.java
@@ -1,22 +1,20 @@
 package org.grobid.core.data;
 
 import org.grobid.core.main.LibraryLoader;
+import org.hamcrest.Matchers;
 import org.junit.Before;
-import org.junit.Ignore;
 import org.junit.Test;
 
 import java.util.Arrays;
 
 import static org.hamcrest.CoreMatchers.is;
 import static org.hamcrest.Matchers.hasSize;
-import static org.junit.Assert.*;
 
 import static org.junit.Assert.assertThat;
 
 import java.io.IOException;
 import java.io.StringReader;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.List;
 
 import javax.xml.parsers.DocumentBuilder;
@@ -31,7 +29,6 @@ import javax.xml.xpath.XPathFactory;
 import org.grobid.core.engines.config.GrobidAnalysisConfig;
 import org.grobid.core.utilities.GrobidProperties;
 import org.junit.BeforeClass;
-import org.junit.Test;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.w3c.dom.Document;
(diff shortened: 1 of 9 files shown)


kermitt2:
I see that there are actually two different `cleanDOI` methods:

https://github.com/kermitt2/grobid/blob/master/grobid-core/src/main/java/org/grobid/core/data/BiblioItem.java#L1007

and

https://github.com/kermitt2/grobid/blob/master/grobid-core/src/main/java/org/grobid/core/utilities/Consolidation.java#L615
(the one modified here)

The second one is not really more specific to CrossRef and both actually normalizes the prefix, so I think we could move everything to `BiblioItem.java` ?

lfoppiano:
Sure! 
I consolidated the method within BiblioItem.java. I also simplified the code and added some tests.

grobid-core/src/test/java/org/grobid/core/data/BiblioItemTest.java
```diff
@@ -172,7 +169,24 @@ public void shouldNotGnerateRawAffiliationTextIfNotEnabled() throws Exception {
...
     @Test
-    public void injectDOI() {
+    public void injectIdentifiers() {
+        BiblioItem item1 = new BiblioItem();
+        item1.setDOI("10.1233/23232/3232");
+        item1.setPMID("pmid");
+        item1.setPMCID("bao");
+        item1.setPII("miao");
+        item1.setIstexId("zao");
+        item1.setArk("Noah!");
```
kermitt2, line 179: Brilliant naming ! :D

kermitt2:
Thanks for the changes, code improvement and the new tests !!

grobid-core/src/test/java/org/grobid/core/data/BiblioItemTest.java
lfoppiano: 😄

lfoppiano:
BTW would we need something like that in biblio-glutton too?

kermitt2:
Not sure, in principle any UTF-8 characters are allowed in the suffix of a DOI, so the composed characters like that too. Obviously it's a character composition mistake error here, but we might have more common diacritics in a DOI. 

It might be more an issue to open in pdfalto, because the character composition should not happen there in principle.

kermitt2:
One possible interesting exercise could be to check in biblio-glutton all the existing DOI and see how many have non-ASCII characters and how many of these are "composed" ones (diacritics).

lfoppiano:
Sure! 

BTW my idea was to have this method called for the DOI supplied to the API. Not at loading time.

kermitt2:
> BTW my idea was to have this method called for the DOI supplied to the API. Not at loading time.

This is how I understand it too, at loading time it would make no sense because it's the valid content that we want to support. The problem is that this PR could break something, when valid diacritics are present in a DOI and cause false negative - so it's interesting to see if such diacritics are actually present or not in CrossRef.

lfoppiano:
Ahh I see 👍

Issue #760, opened 2021-05-26
Publication date potentially corrected from the consolidation service seems to be lost in a limbo 


lfoppiano:
Related to #759. 
I tried to solve the problem by enabling the consolidation, however, the correct publication date was not retrieved correctly. 

The BiblioItem object from the consolidation service had: 
 - publicationDate = null 
 - normalisedPublicationDate with the correct value

(I wonder if this is the correct behaviour)

however, in the BiblioItem, after the `correct(...)` method is invoked, the normalisedPublication date is overridden with the first raw publication date: 

```java
// normalization of dates
                if (resHeader != null) {
                    if (resHeader.getPublicationDate() != null) {
                        List<Date> dates = parsers.getDateParser().processing(resHeader.getPublicationDate());
                        // most basic heuristic, we take the first date - to be
                        // revised...
                        if (dates != null) {
                            if (dates.size() > 0) {
                                resHeader.setNormalizedPublicationDate(dates.get(0));
                            }
                        }
                    }
```

This will inevitably lose the good normalised date from the consolidation service. I'm not sure what is the correct behaviour, I supply this issue mostly for information as I assume there many different use cases 😉

kermitt2:
Good catch Luca !! 
It's a bug yes, we should not normalize the publication date if `(resHeader.getNormalizedPublicationDate() != null)` already. 

> (I wonder if this is the correct behaviour)

yes it's the correct behavior, crossref record has only a normalized publication date (no "raw" publication date).

lfoppiano:
Just for reference, this problem is a bit wider than just in random documents. I have more cases, it seems generally due to document segmentation but happening also with the old pdfalto version. 
With the old pdfalto version on Mac and consolidation enabled, this document: 

https://aip.scitation.org/doi/10.1063/1.3238557 

returns: 

```
<date type="published" when="1036010370">10360 10370 10380 10390 10400</date>
```

when processing fulltext.

PR #761, opened 2021-05-27
override normalised date only when is empty

lfoppiano:
I've implemented a fix for #760. I leave the PR as a draft as it might require a couple of iterations... 

Right now the normalised date is updated only if it's empty. We could add some additional checks from the date parser to check that the date is actually within the minimal requirement to be valid. Is there already code for doing that? 

The output is: 
```
<date type="published" when="2008-05-28">2775 2773 2773 0 2 3 2968 2947 2947</date>
```
@kermitt2 let me know if you have any remark.


kermitt2:
Right now, the only way to have the normalised date not empty (and so a date that should not be normalized) is after successful consolidation. So in this case, I think it's better to also update non-normalized date during consolidation (via the method `Utilities.dateToString()` for example), to have something like that:

```xml
<date type="published" when="2008-05-28">2008-05-28</date>
```

About the case where we have a list of normalized dates returned from a raw extracted date field, I think the problem is not that much impossible dates (I don't remember really having seen some), but rather several incomplete dates that could be combined in one. The way to study that could be to process some thousand of PDF, get all the extracted dates and see where are the problems - but it's quite a lot of time and we probably don't want to do that now with this fix?

lfoppiano:
> Right now, the only way to have the normalised date not empty (and so a date that should not be normalized) is after successful consolidation. So in this case, I think it's better to also update non-normalized date during consolidation (via the method `Utilities.dateToString()` for example), to have something like that:
[..]

That method requires a `java.util.Date`, while we have `org.grobid.data.Date`. I wonder if there isn't any method to transform the Grobid Date to its ISO format. 


> About the case where we have a list of normalized dates returned from a raw extracted date field, I think the problem is not that much impossible dates (I don't remember really having seen some), but rather several incomplete dates that could be combined in one. The way to study that could be to process some thousand of PDF, get all the extracted dates and see where are the problems - but it's quite a lot of time and we probably don't want to do that now with this fix?

I guess it's better to limit this PR to just the fix 😉 
Since we have already the normalised date separated into the year, month and day, if they are not null (or equal to -1), we could just check that year has 4 digits, months and days has 2 digits. We could enforce the intervals for days and months when they are available but I guess that would be a bonus. 

In bug #760 I got a year of 5 digits for example, in which case I have preferred to have none.

kermitt2:
> That method requires a java.util.Date, while we have org.grobid.data.Date. I wonder if there isn't any method to transform the Grobid Date to its ISO format.

Ah yes, I've forgotten the two Date objects in Grobid! The Grobid normalized Date is serialized into ISO format in the TEI. However, I think it would be easier to apply well-formedness criteria to the Grobid object rather than considering java.util.Date, because days and months can often be unspecified (and it does not mean it's first day of the month or first month of the year). 

The `DateParser.normalize()` is the place where date normalization takes place - when I wrote it I didn't apply well-formedness criteria at this stage. I think my idea was to leave further constrains to the applications using Grobid, but for years with 5 digits or far in the future, it's for sure good to reject it in GROBID directly.

lfoppiano:
WIth the last commit, I've improved a bit the efficiency (avoiding parsing the date if there is already a normalised date) and we obtain the ISO version: 

```
<date type="published" when="2008-05-28">2008-05-28</date>
```

For that, I created a new static method in the TEIFormatter.java, and I modified the formatter to use it. 

I've added a few code improvement to avoid NPE in the various IFs and I've removed a hardcoded date (year 2025) in the `DateParser.normalise()` method. 

Maybe we could define a new method for simple data validation that can be used at the end of the `DateParser.normalise()`?

lfoppiano:
I think this PR can be reviewed

grobid-core/src/main/java/org/grobid/core/engines/DateParser.java
```diff
@@ -209,65 +218,76 @@ else if ((date.getDayString().charAt(date.getDayString().length() - 1) == '-')
 
     public static final Pattern[] months = {jan, feb, mar, apr, may, jun, jul, aug, sep, oct, nov, dec};
 
-    public void normalize(Date date) {
+    public Date normalize(Date date) {
+        return normalize(date, false);
+    }
+    
+    
+    public Date normalize(Date date, boolean validate) {
```
de-code, line 226: minor: I'd suggest to have a separately named method instead,e.g. `normalizeDateAndValidate`. Otherwise one would need to be familiar with the function in order to understand what the true in `normalize(..., true)` means.

grobid-core/src/main/java/org/grobid/core/engines/DateParser.java
```diff
@@ -82,17 +91,17 @@ public DateParser() {
...
-                                (!s1.equals(lastTag) && !lastTag.equals("I-<year>"))
+                                (!s1.equals(lastTag) && !"I-<year>".equals(lastTag))
                                 ) {
                             // new date
                             if (date.isNotNull()) {
                                 if (dates == null)
                                     dates = new ArrayList<Date>();
-                                normalize(date);
-                                dates.add(date);
+                                Date normalisedDate = normalize(date, true);
```
de-code, line 103: a bit of British English to honour the queen? :D

grobid-core/src/main/java/org/grobid/core/engines/DateParser.java
```diff
@@ -209,65 +218,76 @@ else if ((date.getDayString().charAt(date.getDayString().length() - 1) == '-')
...
             }
         }
 
         // if we don't have day and month, but a year with 8 digits, we might have a YYYYMMDD pattern
-        if (date.getDay() == -1 && date.getMonth() == -1 && date.getYear() != -1 && date.getYear() > 19000000 && date.getYear() < 20251231) {
+        int maxYear = Calendar.getInstance().getWeekYear() + 4;
+        if (date.getDay() == -1 && date.getMonth() == -1 && date.getYear() != -1 && date.getYear() > 19000000 && date.getYear() < maxYear * 10000+1231) {
             int yearPart = date.getYear() / 10000;
-            if (yearPart > 1900 && yearPart < 2025) {
+            if (yearPart > 1900 && yearPart < maxYear) {
```
de-code, line 290: time flies, I guess when that line was written we thought there would be flying cars all over by now.

grobid-core/src/main/java/org/grobid/core/engines/DateParser.java
```diff
@@ -288,16 +308,59 @@ public void normalize(Date date) {
...
+    /**
+     * Simple and loose date validation, checking: 
+     *  - the year has not more than 4 digits
+     *  - the month and day has not more than 2 digits 
+     *  
+     *  Assuming that incomplete dates of any form and nature can pass by here, only the information that are "out of bounds" will be reverted.
+     *  
+     * @return the date where invalid information are removed or reverted
+     */
+    public static Date postValidate(Date originalDate) {
```
de-code, line 337: minor: I guess in some sense it feels more like "cleaning" than validating. Although loose terms anyway.

de-code:
Many parts seem to read better. I don't feel fully qualified to review what the fix is actually about, as I haven't really used the consolidation. (It does seem to make sense in general)

grobid-core/src/main/java/org/grobid/core/engines/DateParser.java
lfoppiano: good point

grobid-core/src/test/java/org/grobid/core/engines/DateParserIntegrationTest.java
```diff
@@ -62,13 +61,16 @@ public void processing_englishStandardDate1_shouldWork() throws Exception {
 
         assertThat(date.getDayString(), is("19"));
         assertThat(date.getMonthString(), is("January"));
-        assertThat(date.getYearString(), is("19 83"));
+
+        // TODO: With the clusteror the space is removed...
```
lfoppiano, line 65: This test is failing because the clusteror removes the spaces when extracting the text of the concatenated tokens

lfoppiano:
As discussed, I've migrated the DateParser to the Clusteror classes. 
Few comments: 
 - I've renamed the various method using `process()` instead of `processing`. It seems more correct to me, howver we can revert it back. I left the original one with the deprecation note, we can revert. 
 - I've kept the `List<String>` because I wasn't feeling too confident to replace it with the `List<LayoutToken>`.  
 - The unit tests are all working except one (see [comment](https://github.com/kermitt2/grobid/pull/761#discussion_r648022774)).

kermitt2:
@lfoppiano Hi Luca! Would it be possible on your side to update the test class for new config to have the branch merging without conflicts? Thanks !

lfoppiano:
I updated the branch. The ParseDate test is fixed. 

There are two tests in the org.grobid.core.engines.SegmentationTest which are failing for me on the macOs.. 
The change at rev a012f6ba9725fdcbdab5e626172aba9a4cd6baf4 seems to be causing the problem for me, but I guess it solved it for you on Linux. 

I wonder whether we have the same pdfalto version in both platforms...

kermitt2:
When running test with latest PR version, I have 16 fails all due to this: 

```
  java.lang.ExceptionInInitializerError
      at org.grobid.core.engines.DateParserTest.setUp(DateParserTest.java:38)
  Caused by: java.lang.NullPointerException
      at org.grobid.core.engines.DateParserTest.setUp(DateParserTest.java:38)
```

grobid-core/src/main/java/org/grobid/core/engines/DateParser.java
```diff
@@ -1,180 +1,214 @@
...
+        super(model);
+    }
+    
     /**
-     * Processing of authors in header
-     */
+     * Deprecated, @Use process(String input)
+     **/
+    @Deprecated
     public List<Date> processing(String input) {
```
kermitt2, line 43: I would keep `processing()` like everywhere else, because why just changing only this one? It's very minor difference and I think better be consistent?

kermitt2 (commented):
When running test with latest PR version, I have 16 fails all due to this: 

```
  java.lang.ExceptionInInitializerError
      at org.grobid.core.engines.DateParserTest.setUp(DateParserTest.java:38)
  Caused by: java.lang.NullPointerException
      at org.grobid.core.engines.DateParserTest.setUp(DateParserTest.java:38)
```

grobid-core/src/main/java/org/grobid/core/engines/DateParser.java
lfoppiano: I left the `processing()` method, but deprecated it so that both can be used and the users will know that at some point will be removed. Since the change was related to the DateParser I did not modify all the parsers in this PR.

lfoppiano:
> When running test with latest PR version, I have 16 fails all due to this:
> 
> ```
>   java.lang.ExceptionInInitializerError
>       at org.grobid.core.engines.DateParserTest.setUp(DateParserTest.java:38)
>   Caused by: java.lang.NullPointerException
>       at org.grobid.core.engines.DateParserTest.setUp(DateParserTest.java:38)
> ```

Sorry, there was a modification that I forgot to commit. It's pushed now.

grobid-core/src/main/java/org/grobid/core/engines/DateParser.java
kermitt2: ok I count on you then to change elsewhere then :D (I would not have the courage for all these changes just for 3 letters ;) ) 
It could be an independent PR where the method is deprecated everywhere?

lfoppiano: OK. Usually the IDE will do most of the job 👍

PR #778, opened 2021-06-24
Training web services

kermitt2:
Makes possible to launch the training of a particular model (`/api/modelTraining`), monitor training advancement (`/api/trainingResult`) and retrieve a model (`/api/model`) via web services (for instance with the a container).

It is coming from another project but to be tested here.


kermitt2:
Everything looks good. Documentation updated.

PR #800, opened 2021-07-19
Accept application/x-bibtex for processHeaderDocument

btut:
I extended the api to get bibtex for the processHeaderDocument service.
Previousely, AFAIK, it was only possible for /api/processReferences and /api/processCitation services.


koppor:
This is a follow-up to https://github.com/kermitt2/grobid/pull/532

kermitt2:
Hello @btut !

Thanks for the PR. Did you look how the BibTeX results for the header metadata look like? 

I vaguely remember that the extracted header fields are not exactly mapped to the same field as for a citation string in the BiblioItem object (because they are handled and normalized differently), so the BibTeX serialization might not work correctly for headers - of course anyway many of the interesting extracted fields will be lost in the very limited BibTeX format as compared to the XML version. 

So given your feedback, we might need to extend/review the `toBibTeX()` method in `BiblioItem` class (which I didn't look at since 7-8 years :) for having this PR working.

btut:
Hi!
> Did you look how the BibTeX results for the header metadata look like?

I checked a handful of pdfs and results seem fine. The only necessary field that seems to be missing is the date. I'll check the toBibTeX method to see what else might be lost.

> of course anyway many of the interesting extracted fields will be lost in the very limited BibTeX format as compared to the XML version

Sure, but I think the goal for someone that needs BibTeX is to cite it. Aside from the date, the BibTeX entry only lacks author details like e-mail and affiliation. There is no need for these details when citing.

> we might need to extend/review the toBibTeX() method in BiblioItem class

I'll have a look!

btut:
Hello again!
I fixed a small problem with the dates in toBibTeX. I noticed that usually the normalized_publication_date field is populated, not the pubblication_date so I used that one and used ISO 8601 formatting (used by BibTeX).

Unfortunately, for many papers that I used for testing Grobid did not detect a date at all. This was the case both for BibTeX and TEI.

btut:
> I'll have a look!

I tested this extensively with many pdf's and it looks good! Could we move on with this PR? We would like to use this feature in JabRef.

grobid-core/src/main/java/org/grobid/core/data/BiblioItem.java
```diff
@@ -1973,9 +1973,12 @@ public String toBibTeX(String id, GrobidAnalysisConfig config) {
             }
             // fullEditors has to be used instead
 
-            // year
-            if (publication_date != null) {
-                bibtex.add("  year = {" + publication_date + "}");
+            // dates
+            if (normalized_publication_date != null) {
+                String isoDate = normalized_publication_date.toISO();
```
kermitt2, line 1978: There's already an ISO normalization method to be re-used: 
https://github.com/kermitt2/grobid/blob/master/grobid-core/src/main/java/org/grobid/core/document/TEIFormatter.java#L108

grobid-core/src/main/java/org/grobid/core/data/Date.java
```diff
@@ -214,4 +214,19 @@ public String toXML() {
...
+        if (month == -1) {
+            return theDate;
+        }
+        theDate += "-" + String.valueOf(month);
+        if (day != -1) {
+            theDate += "-" + String.valueOf(day);
+        }
+        return theDate;
+    }
+
```
kermitt2, line 231: Re-use: https://github.com/kermitt2/grobid/blob/master/grobid-core/src/main/java/org/grobid/core/document/TEIFormatter.java#L108
-> it manages correctly the 0-digit for match the ISO pattern.

kermitt2 (changes requested):

kermitt2:
Hello @btut !

I submitted a review for changes: 1) reuse the existing ISO date normalization (more complete - it has been changed/moved recently, so you might need to merge your PR branch with the current master) and 2) review the returned MIME type. 

Apart from that, I've seen one bug, when the surname is incorrectly recognized as forname, we have a `null` in the BibTeX: 

```
curl -X POST -H "Accept: application/x-bibtex" -d "citations=Griff, Expert. Opin. Ther. Targets (2002) 6(1): 103-113" localhost:8070/api/processCitation

@article{-1,
  author = {null, Griff},
  journal = {Expert. Opin. Ther. Targets},
  year = {2002},
  pages = {103--113},
  volume = {6},
  number = {1}
}
```
```
lopez@work:~/grobid_client_python$ curl -X POST -H "Accept: " -d "citations=Griff, Expert. Opin. Ther. Targets (2002) 6(1): 103-113" localhost:8070/api/processCitation
<biblStruct >
	<analytic>
		<title/>
		<author>
			<persName><forename type="first">Griff</forename></persName>
		</author>
	</analytic>
	<monogr>
		<title level="j">Expert. Opin. Ther. Targets</title>
		<imprint>
			<biblScope unit="volume">6</biblScope>
			<biblScope unit="issue">1</biblScope>
			<biblScope unit="page" from="103" to="113" />
			<date type="published" when="2002" />
		</imprint>
	</monogr>
</biblStruct>
```

We might want to output no author at all in the BibTeX in this case? 

The rest look very good indeed, everything is well mapped as expected!

btut:
Thanks for your review, @kermitt2!

> Apart from that, I've seen one bug, when the surname is incorrectly recognized as forname, we have a null in the BibTeX

Great catch!

> We might want to output no author at all in the BibTeX in this case?

I think just outputting the firstname is better as it preserves that information at least. I implemented that in 2a69fa4, let me know if you disagree!

kermitt2:
Thanks a lot @btut for the changes !

kermitt2 (approved):

btut:
Thanks for your great work! Grobid seems to work very well!

kermitt2:
Too fast (as often), the change is actually not passing the tests because of the way dates are now outputted in the bibtex: 

- if I understand well `date` in BibTeX is always ISO normalized?

- for more "raw date", BibTeX uses `year`, for instance for this: 
```
Kolb, S., Wirtz G.: Towards Application Portability in Platform as a Service Proceedings of the 8th IEEE International Symposium on Service-Oriented System Engineering (SOSE), Oxford, United Kingdom, April 7 - 10, 2014.
```
we dont' have a normalized date, because of the day range, so we would have something like this in the bibtex: 

```
year = {April 7 - 10, 2014},
```

- we can have too something like that:

```
year = {2014},
month = {April},
day = {7 - 10},
```

and apparently the support of these bibtex flavors depend on the style?

kermitt2:
Is something like this acceptable, with both `date` and `year` to reflect normalized versus raw extraction: 

```
curl -X POST -H "Accept: application/x-bibtex" -d "Kolb, S., Wirtz G.: Towards Application Portability in Platform as a Service Proceedings of the 8th IEEE International Symposium on Service-Oriented System Engineering (SOSE), Oxford, United Kingdom, April 7 - 10, 2014." localhost:8070/api/processCitation
```

```
@inproceedings{-1,
  author = {Kolb, S and Wirtz, G},
  date = {2014},
  year = {April 7 - 10, 2014},
  address = {Oxford, United Kingdom}
}
```

btut:
I added some thoughts in #807.
> Is something like this acceptable, with both date and year to reflect normalized versus raw extraction

I think this would be a good way to go as it gives all possible data and the user can decide on what to use. I would go the other way around though (split `normalized_publication_date` into `year`, `month` and `day` and place the non-normalized date in the `date` field), because `year` should be numeric.

koppor:
TLDR:


>     * we can have too something like that:
> 
> 
> ```
> year = {2014},
> month = {April},
> day = {7 - 10},
> ```

Please try to output following:

```bibtex
year = {2014},
month = 4,
date = {2014-04-07/2014-04-10},
```

Long answer:

>     * if I understand well `date` in BibTeX is always ISO normalized?

This is done by the bibtex-variant "biblatex". See https://ftp.rrzn.uni-hannover.de/pub/mirror/tex-archive/macros/latex/contrib/biblatex/doc/biblatex.pdf

"iso8601-2 Extended Format specification level 1", which is a "yes" to your answer.

Nevertheless, date ranges can be specified:

grafik

However, normal BibTeX does not use that, it only uses `year` and `month`. In all cases, the `year` field should be filled to be compatible with most `.bib` processors out there. - The `date` field is more an optional thing and could be left out.

> we dont' have a normalized date, because of the day range, so we would have something like this in the bibtex:
> 
> ```
> year = {April 7 - 10, 2014},
> ```

This should not happen. :)

> and apparently the support of these bibtex flavors depend on the style?

Yes, they do. However, `year` and `month` is (nearly) always supported.

Issue #632, opened 2020-09-01
Error case for the segmentation model 


lfoppiano:
I have this PDF (https://www.nature.com/articles/s41598-020-58065-9.pdf) and I notice some issue in processing it. 

It seems that everything goes in the body section from the segmentation parser: 
[segmentation.txt](https://github.com/kermitt2/grobid/files/5154961/segmentation.txt)

This is the output from pdfalto: 
[3SHt4RC7GX.xml.txt](https://github.com/kermitt2/grobid/files/5154978/3SHt4RC7GX.xml.txt)

I also noticed that this pdf contains a lot of characters that are not encoded correctly, maybe can be used as a test case for pdfalto?

kermitt2:
It looks like it only concerns the header zone recognition, the rest of the segmentation labels are more or less good. But as it's the header, it is very visible of course. Normally Nature open access publication are CC-BY so we could add some training data and fix the error case like that. At some point, the segmentation will need again some work with better features, it works very badly when special characters/equations are around, but we need more training data available to make a meaningful job on this.

lfoppiano:
This was implemented in #951

PR #813, opened 2021-08-12
Add training data for one special error case 

lfoppiano:
This PR provides the training data to improve the recognition related to error case #632. I've added in the guidelines a couple of additional sentences related to the author contribution as discussed. 

Where I was not sure, I commented directly in the XML files.


kermitt2:
Hello Luca ! 
Would it be possible to send me the PDF for these training cases?

lfoppiano:
This is just one document. 
The PDF was linked in the issue: https://www.nature.com/articles/s41598-020-58065-9.pdf

kermitt2:
Raw header file need to be regenerated after retraining of the segmentation model, because some content has been added in the segmentation file. 

I think it would be better here to roll-back the changes in the guidelines because they will conflict will most recent ones done in other place and they are redundant.

kermitt2:
It might be appropriate to merge this PR with branch `feature/data-availability-statement` first and not with master, because the changes relatively to training in different branches here make things difficult to follow and update in a consistent manner (the raw files and model in particular).

lfoppiano:
Thanks for the review! I close this and add these files (without guidelines) directly in the `feature/data-availability-statement`.

PR #814, opened 2021-08-12
Revise BibTeX date output

btut:
Follow up to #761 and #807.

Date is now given as:

- If date is normalizable:
  - ISO string in date field
  - year integer in year field
  - month integer in month field
  - day integer in day field
- If date is not normalizable:
  - raw date-string in year field

### DIFF
--- a/grobid-core/src/main/java/org/grobid/core/data/BiblioItem.java
+++ b/grobid-core/src/main/java/org/grobid/core/data/BiblioItem.java
@@ -1990,9 +1990,17 @@ public class BiblioItem {
                 if (isoDate != null) {
                     bibtex.add("  date = {" + isoDate + "}");
                 }
-            }
-            if (publication_date != null) {
-                bibtex.add("  year = {" + publication_date + "}");
+                if (normalized_publication_date.getYear() >= 0) {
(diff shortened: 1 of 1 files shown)


grobid-core/src/main/java/org/grobid/core/data/BiblioItem.java
```diff
@@ -1990,8 +1990,16 @@ public String toBibTeX(String id, GrobidAnalysisConfig config) {
...
                 }
-            }
-            if (publication_date != null) {
+                if (normalized_publication_date.getYear() >= 0) {
+                    bibtex.add("  year = {" + normalized_publication_date.getYear() + "}");
+                }
+                if (normalized_publication_date.getMonth() >= 0) {
+                    bibtex.add("  month = {" + normalized_publication_date.getMonth() + "}");
+                }
+                if (normalized_publication_date.getMonth() >= 0) {
```
kermitt2, line 1999: it should be `normalized_publication_date.getDay()` here, corrected when added to #807

btut:
Thanks for the fix and the merge! This PR is not needed anymore.

PR #831, opened 2021-09-07
fix deserialised consolidation day

lfoppiano:
Minor bugfix. The day (String only) in the consolidated date was not correctly assigned from the response.


kermitt2:
Thanks !

PR #832, opened 2021-09-08
training: small fixes to name (citation) data

bnewbold:
When skimming over TEI training data (to see how diverse the source data is), I noticed what seem to be two small labeling errors.

I checked over the rest of this file (author.tei.xml), but didn't inspect the rest of the citation corpus carefully.


kermitt2:
Thanks you !!

PR #833, opened 2021-09-10
change base image of grobid docker for delft

lfoppiano:
this should solve the problem of using the gpus with docker

### DIFF
--- a/Dockerfile.delft
+++ b/Dockerfile.delft
@@ -61,7 +61,7 @@ RUN rm -rf grobid-source
 # -------------------
 
 # use NVIDIA Container Toolkit to automatically recognize possible GPU drivers on the host machine
-FROM nvidia/cuda:10.1-base
+FROM FROM tensorflow/tensorflow:1.15.5-gpu
 CMD nvidia-smi
 
 # setting locale is likely useless but to be sure


lfoppiano:
According to my tests the GPU memory is filled up correctly 

image


The image for testing is pushed with name `lfoppiano/grobid:test`

lfoppiano:
I can confirm it works also with derived docker images ;-)

lfoppiano:
@kermitt2 another aspect that I would be interested to test is to calculate the maximum batch size given a fixed max length size for a certain amount of GPU memory, I guess this could be done with delft via a python script?

kermitt2:
> another aspect that I would be interested to test is to calculate the maximum batch size given a fixed max length size for a certain amount of GPU memory, I guess this could be done with delft via a python script?

It depends on the model architecture. So I think there is no way in general to know in advance, but it might be related to the number of parameters of the model.

lfoppiano:
> It depends on the model architecture. So I think there is no way in general to know in advance, but it might be related to the number of parameters of the model.

Sorry I meant by setting also the architecture

kermitt2:
> Sorry I meant by setting also the architecture

It sounds complicated to do in general for DeLFT, but for Grobid we can't know in advance the number of models that would be running on GPU so automating this is probably not doable.

lfoppiano:
> > Sorry I meant by setting also the architecture
> 
> It sounds complicated to do in general for DeLFT, but for Grobid we can't know in advance the number of models that would be running on GPU so automating this is probably not doable.

Yes sorry, indeed it makes no sense if more models are used in parallel. 

What I'm not sure of is that JEP uses just one thread, right? So,  does it means that different models cannot be called in parallel?

kermitt2:
In the current settings, all the DL models together are using an Executor of size 1, so different models can be loaded and called in parallel, but only one will run at the same time. The other are blocked while the run of the current one is done. 
The motivation is that, if I am not wrong, one single model inference is done at the same time on one GPU (there is a kind of lock of the GPU managed by TensorFlow), so one JEP thread is enough. However, the JEP instance will indeed load and initialize several models if requested. 
 
If you look at https://github.com/kermitt2/grobid/blob/master/grobid-core/src/main/java/org/grobid/core/jni/JEPThreadPool.java it's possible to manage a thread of pools with several JEP instances (one JEP instance is bounded by the thread that initialized it), so it should possible in theory to have multiple models running in different threads. But I think because I've never been able to use more than one GPU, I've never been able to really test the usage of parallel inference on multiple GPU further.

lfoppiano:
@bnewbold Could you test it with CUDA 11.x and let me know whether it works? It should work fine with CUDA 10.1 / 10.2

kermitt2:
My current tests: 
- CUDA 9 -> not working, but not a problem
- fallback to CPU when no GPU -> working

For CUDA 10.*, I still have a process running on the machine... I will also try to build on a machine without GPU and test on machine with GPU to be complete.

bnewbold:
There is a typo in this change (`FROM FROM`), resulting in:

```
Error response from daemon: dockerfile parse error line 64: FROM requires either one or three arguments
```

lfoppiano:
> There is a typo in this change (`FROM FROM`), resulting in:
> 
> ```
> Error response from daemon: dockerfile parse error line 64: FROM requires either one or three arguments
> ```

oh sorry about that... 

You could try directly the image `lfoppiano/grobid:test`

kermitt2:
> There is a typo in this change (FROM FROM), resulting in:

@lfoppiano gasp you couldn't see my comment right? How can I make those things more visible? 

Screen Shot 2021-09-14 at 06 13 17

lfoppiano:
> > There is a typo in this change (FROM FROM), resulting in:
> 
> @lfoppiano gasp you couldn't see my comment right? How can I make those things more visible?
> 
> Screen Shot 2021-09-14 at 06 13 17

oh... I did not see it.. 
I suppose that after the comment you have to "start review"? 
I'm not sure but the Github workflow for review is too clumsy ....

kermitt2:
> I suppose that after the comment you have to "start review"?

it's what I did :(

> I'm not sure but the Github workflow for review is too clumsy ....

yes... not sure how to do it now.

bnewbold:
I started testing on two VMs, but got interrupted and had to tear them down. If I have time tomorrow i'll give it a try.

kermitt2:
It works for me too with CUDA 10.1 

```
|===============================+======================+======================|
|   0  GeForce GTX 105...  Off  | 00000000:01:00.0  On |                  N/A |
| 45%   37C    P8    N/A /  75W |   3875MiB /  4036MiB |     17%      Default |
+-------------------------------+----------------------+----------------------+
                                                                               
+-----------------------------------------------------------------------------+
...
|    0     22139      C   java                                        3335MiB |
```

kermitt2:
I've updated the documentation and the image on DockerHub... this is ready to merge I think.

lfoppiano:
Great!

PR #838, opened 2021-10-05
prefer author meta from consolidation , crossref is considered more r…

Aazhar:
Using some editors formats, extracting ORCID using PDF annotations is not reliable in this case if consolidated , then prefer the crossref metadata for the authors ORCIDS.


kermitt2:
Thank you @Aazhar  !

Issue #848, opened 2021-11-03
TEI-XML output: un-escaped ampersand ('&') in ISSN <idno> tag


bnewbold:
When parsing this raw citation string with `processCitation` or `processCitationList`:

```
Maheswari, S., Vijayalakshmi, C.: Optimization Model for Electricity Distribution System Control using Communication System by La-grangian Relaxation Technique. CiiT International Journal of Wireless Communication 3(3), 183–187 (2011) (Print: ISSN 0974 – 9756 & Online: ISSN 0974 – 9640)
```

The output XML contains a parsed ISSN `<idno>` tag like:

```
<idno type="ISSN">0974 - 9756 & Online: 0974 - 9640</idno>
```

The ampersand character is not escaped, which makes the entire XML snippet invalid. It should be a `&amp;`. If the raw citation is returned, the `<note>` tag is escaped correctly:

```
<note type="raw_reference">Maheswari, S., Vijayalakshmi, C.: Optimization Model for Electricity Distribution System Control using Communication System by La-grangian Relaxation Technique. CiiT International Journal of Wireless Communication 3(3), 183–187 (2011) (Print: ISSN 0974 – 9756 &amp; Online: ISSN 0974 – 9640)</note>
```

It looks like the problem is missing `TextUtilities.HTMLEncode()` around here: https://github.com/kermitt2/grobid/blob/master/grobid-core/src/main/java/org/grobid/core/data/BiblioItem.java#L2646

for both ISSN and ISSNe outputs. I can provide a patch if it would be helpful

The ISSN also isn't validated, and includes two ISSNs, but that is a separate more complicated issue.

bnewbold:
From a quick scan, it looks like these other fields are also not escaped in the same file; this might not be an exhaustive list:

- `HALFile` (`url`)
- `HALid` (`uri`)
- <keywords scheme="hal"> (`subject`)
- `<keywords type="subject-headers">` (`keywords1`)
- `<idno>` (`uri`) when uri startsWith "http://hal."
- `town` and `country` in the context of `<meeting>` (line 2437)

kermitt2:
Yes we should escape all these fields too, because garbage in/garbage out ! 

> The ISSN also isn't validated, and includes two ISSNs, but that is a separate more complicated issue.

[...]

PR #852, opened 2021-11-09
Escape more identifiers, including ISSN

bnewbold:
Closes https://github.com/kermitt2/grobid/issues/848


kermitt2:
Thank you @bnewbold for these fixes and the tests. 
I will check other TEI serialization pieces for the same possible issue.

Issue #845, opened 2021-10-25
biblStruct XML: multiple editors, single <editor> tag


bnewbold:
Current (0.7.x) GROBID output for multiple authors results in multiple `<author>` tags, each with a `<persName>`. But multiple editors results in a single `<editor>` tag with multiple nested `<persName>`.

I'm not an expert at TEI-XML, but a quick skim of [editor tag docs](https://tei-c.org/release/doc/tei-p5-doc/en/html/examples-editor.html) make it look like multiple `<editor>` tags would be the norm.

The relevant source code is https://github.com/kermitt2/grobid/blob/master/grobid-core/src/main/java/org/grobid/core/data/BiblioItem.java#L2382

Here is an example citation string:

```Uhlířová, K., Drumbl, M. (forthcoming 2020): Actors and Law Making in International Environmental Law, in M. Fitzmaurice, M. Brus and P. Merkouris (eds.): The Research Handbook on International Environmental Law. Cheltenham: Edward Elgar Publishing. 50 p.```

and TEI-XML output from `processCitation` (via https://cloud.science-miner.com/grobid/ just now):

```xml
<biblStruct >
    <monogr>
        <author>
            <persName>
                <forename type="first">K</forename>
                <surname>Uhlířová</surname>
            </persName>
        </author>
        <author>
            <persName>
                <forename type="first">M</forename>
                <surname>Drumbl</surname>
            </persName>
        </author>
        <title level="m">Actors and Law Making in International Environmental Law</title>
        <editor>
            <persName>
                <forename type="first">M</forename>
                <surname>Fitzmaurice</surname>
            </persName>
            <persName>
                <forename type="first">M</forename>
                <surname>Brus</surname>
            </persName>
            <persName>
                <forename type="first">P</forename>
                <surname>Merkouris</surname>
            </persName>
        </editor>
        <meeting>
            <address>
                <addrLine>Cheltenham</addrLine>
            </address>
        </meeting>
        <imprint>
            <publisher>Edward Elgar Publishing</publisher>
            <date type="published" when="2020" />
            <biblScope unit="volume">50</biblScope>
        </imprint>
    </monogr>
    <note>The Research Handbook on International Environmental Law</note>
</biblStruct>
```

I can prepare a patch if that is helpful.

kermitt2:
Hi @bnewbold !

[...]

PR #853, opened 2021-11-09
TEI: multiple editors each get <editor> around <persName>

bnewbold:
Closes: https://github.com/kermitt2/grobid/issues/845

Tested with the string mentioned in the issue, but didn't add any unit test coverage.


kermitt2:
Thanks a lot !

PR #857, opened 2021-11-12
visualizer: use arxiv.org/abs/<id> instead of arxiv.org/<id>

bnewbold:
Small tweak to use a more direct URL for arxiv.org preprints


kermitt2:
This is working much better indeed, I think because there is a redirection https://arxiv.org/abs/arxiv:0910.2050 -> https://arxiv.org/abs/0910.2050 and I am not removing the `arxiv:` prefix in the url.

Issue #275, opened 2018-01-09
arXiv identifiers not extracted


dksanyal:
Hi, I found that arXiv identifiers are not extracted from PDF files in most cases (both in the online version 
 and in our local build). I have uploaded 2 example PDFs where this problem occurs. 
(I have seen only a very few cases where it succeeds.) Kindly let me know if GROBID supports extraction of arXiv ids, and, if not, is there any plan to support it in the near future
[1801.00857.pdf](https://github.com/kermitt2/grobid/files/1615064/1801.00857.pdf)
[1801.02609.pdf](https://github.com/kermitt2/grobid/files/1615065/1801.02609.pdf)

.

kermitt2:
Hello, arXiv identifiers are well supported by the bibliographical reference models (nearly 2000 annotated examples haven been added with arXiv ids), however the header model has not been yet updated to support them similarly. So arXiv present in the header of the PDf will in general not been extracted for the moment. 

I've planned to add that when I will update the header model, which is long over due... as it is quite a lot of work (training data will need to be updated too), I need to find enough free time to launch the effort - it should done in the first half of this year, so rather mid-term future.

dksanyal:
Thank you Sir for your response and for meticulously maintaining such a complex tool for the community. 
We hope to see the updated header model soon!

philgooch:
I find regex extraction for well-defined identifiers such as arXiv ids and DOI works well, rather than training a model to detect them.

For modern arXiv ids, something like `Pattern.compile("(?i)arXiv:\\d{4}[.]\\d{4,5}(v\\d+)?")` works well.

`getAllBlocksClean()` in the Grobid Document class will give you the raw text of the document, you can then restrict the arXiv id search to the first few lines of the document to reduce the risk of picking up an arXiv id from the bibliography.

kermitt2:
hello @philgooch !

DOI detection works exactly like this in the header since something like 4-5 years.

This is working very well for very discriminant identifiers indeed. 

The advantage of integrating these regex in the ML model as feature (there is no ML model trained specifically for any type of identifier in GROBID) like in the citation parser, is that it helps also the prediction of other structures around, it makes the process more robust to PDF noise and it could work for more ambiguous identifiers like ISSN - so it's more general. For the citations and arxiv ids, this approach was better performing than only using a regex on the citation string independently from the CRF (but not a lot better neither if I remember well...). For the header, we will see!

kermitt2:
[...]

PR #858, opened 2021-11-13
expand scope of arxiv identifier matcher, and fix some training data annotations

bnewbold:
The current arxiv.org identifier matcher regex requires an "arXiv:" prefix. This misses some un-ambiguous old-style identifiers in short citations, like these examples;

```
B.A. Dobrescu, hep-ph/9510424.
K.R. Dienes, C. Kolda and J. March-Russell, hep-ph/9610479.
S.P. Martin, hep-ph/9608224.
```

This PR extends the regex to allow these cases, by more conservatively matching the "section" prefix, and not allowing whitespace, to reduce the risk of matching on random garbled text. The above examples are added as citation training data.

I also noticed a handful of existing training citations which were missing `type="arxiv"` on the `<idno>` tag, so I added those.

There is some possible ambiguity about `arxiv` vs. `arXiv` in the TEI data: GROBID output has the capitalized X, while the training data is lower-case. Also, while all (or almost all) the citation training data has these annotations, none of the header `<idno>` annotations do. Not sure if it would be helpful to add those.

This is related to https://github.com/internetarchive/fatcat/issues/84 and to https://github.com/kermitt2/grobid/issues/275


elonzh:
FYI, https://github.com/mattbierbaum/arxiv-public-datasets/blob/master/arxiv_public_data/regex_arxiv.py

kermitt2:
@bnewbold Thanks a lot for the improvement ! I think there are small issues in the new regex and I propose some changes in the review after some tests, if you could have a look?

bnewbold:
@kermitt2 thanks for reviewing.

I don't see any proposed changes or review notes here on github. Maybe I am missing something?

kermitt2 (changes requested):

kermitt2:
Ok sorry I think I did not submit the review, can you see my comments now?

PR #870, opened 2021-12-07
[WIP] Support for biblio-glutton v0.2

kermitt2:
Here are the changes to work with the new version of biblio-glutton.


lfoppiano:
I would revert c8f1a4d back so that it uses the shipped gradle instead of the provided one by circleci. 
Also the comments on the jacoco plugin will be helpful to track the coverage and see if the tests are failing directly in the circleci console.

kermitt2:
@lfoppiano yes! I don't remember exactly why I changed it, I think I was stuck with the gradle test, but I wanted to make the changes for glutton 0.2. I make clearer that this WIP.

lfoppiano:
OK. Now it's good. We can see the tests results directly on circleci. 

https://92-5797013-gh.circle-artifacts.com/0/build/reports/jacoco/codeCoverageReport/html/index.html

image

Of course also coveralls provides the same informations :-)

Issue #688, opened 2021-01-02
Setting JEP native library path for java >10


kermitt2:
I am opening a specific issue for this, although it was mentioned #603. 

The JEP instance will load itself the native JEP library, but for this to happen we need to add the path to this native library in the `java.library.path`. So far it was done dynamically, via the user `usr_paths` field where we appended the JEP native path, and only if some Deep Learning model was required by the config. Unfortunately `usr_paths` could not be accessed any more as Field via the ClassLoader from Java 11 for security reason. 

A possible solution is to use MethodHandler as explained [here](https://stackoverflow.com/a/59468135) but it seems not working then for java <8 and stops working again for java 15 because `usr_paths` cannot be accessed via the ClassLoader any more at all (`Caused by: java.lang.NoSuchFieldError: usr_paths`). 

So adding new paths for native libraries at runtime in Java seems a dead end. 

The alternative would be to add the JEP native path when launching the JDK via Gradle, with something like this:
```
    systemProperty "java.library.path", file("grobid-home/lib/**os-arch***/").absolutePath
```

If JEP is not used, the library won't be loaded and this path won't be used anyway, so it's not a problem to add the path every time at launch.

For this, we need to identify the right OS info from Gradle (see maybe [here](https://www.jetbrains.com/help/idea/setting-up-jni-development-in-gradle-project.html#adding_jni_support) for an example). 

Note: It does not affect the docker image which will always work fine. In the case of the Docker image, the right JEP is in the system library path because it is fully installed on the image, and we are not using the embedded native JEP library at all.

kermitt2:
Limiting the settings to linux 64, so far I didn't manage to get anything working here via Gradle. 

It's possible to add the system property to tasks like this:

```
task(jatsEval, dependsOn: 'classes', type: JavaExec, group: 'modelevaluation') {
        main = 'org.grobid.trainer.evaluation.EndToEndEvaluation'
        classpath = sourceSets.main.runtimeClasspath
        args 'nlm', getArg('p2t', '.'), getArg('run', '0'), getArg('fileRatio', '1.0')
        jvmArgs '-Xmx3072m'
        systemProperty "java.library.path", file("../grobid-home/lib/lin-64/jep").absolutePath 
    }
```

But it overrides the normal `java.library.path` and fails then to make JEP working as the other library paths are discarded. I didn't find how to extend the path. 

The following also fails for the same reason:

```
[...]

Issue #689, opened 2021-01-02
JEP embedded native library


kermitt2:
Currently the JEP native libraries coming with grobid depend simply on the OS and arch. 32/64 bits, but this is not enough because this built library also depends on the version of the JRE used to run Grobid and the version of python installed locally (to run DeLFT). If we want to still embed JEP in GROBID to have the Deep Learning models working "out-of-the-box" we will need to add native libraries for all this variants, or to limit the supported versions of Java and Python. 

Right now only the native JEP library for Java 8 and python 3.6 is shipped with Grobid for Lin64 and MacOS.

de-code:
Just some feedback from my side.

I am using [sciencebeam-trainer-delft](https://github.com/elifesciences/sciencebeam-trainer-delft).
I am using regular `pip` to install dependencies, including `jep`. The version of `jep` needs to match the one used in GROBID.
Installing `jep` will also build the binary, using the already installed default Java version (JDK).
I am then making sure, that the path to the JEP binary is in `LD_LIBRARY_PATH` (e.g. `LD_LIBRARY_PATH=venv/lib/python3/site-packages/jep`).

That has worked relatively smoothly for me. The pre-build JEP binary within GROBID was not helpful so far for me due to the dependencies that you mentioned.

When deployed, I am using the Docker image built by [sciencebeam-trainer-delft](https://github.com/elifesciences/sciencebeam-trainer-delft). That obviously contains everything, including Java, Python and the JEP binary.

i.e. Docker is a way of providing those things together. There are other formats that could work similar, e.g. `AppImage` for Linux.

When installing `jep` via pip it currently does require the JDK to be installed. I guess it could have provided pre-build binaries. It feels like attempting to include all supported JEP versions in GROBID is something that would rather fit into the scope of JEP. But there are probably too many variables to make that smooth.

PR #921, opened 2022-06-01
Set native path with gradle

kermitt2:
See #688

As dynamically loading native libraries is not possible with JDK > 10 (or with complicated hacks depending on the JVM version), this PR tries to set the `java.library.path` at launch via Gradle. 

The current version covers lin64 only (service, training, evaluation)

To do:
- make it [OS specific](https://www.jetbrains.com/help/idea/setting-up-jni-development-in-gradle-project.html#adding_jni_support)
- update batch command line, or create Gradle exec task?
- the loading of libraries from the python environment config was dynamic too, so it won't be working without similar change, see https://github.com/kermitt2/grobid/blob/4757b0b3c0c14be7905bc74191a6bdfcdd0421ac/grobid-core/src/main/java/org/grobid/core/main/LibraryLoader.java#L159


lfoppiano:
I've made some small progress (on the macOs). 

In brief: 
- I've created the task called `run-server` and I've extended the `library path` using the conda environment. 
- I've removed the libjep.so from the mac os library path because, depending how it was built might require a python version that is not installed. If we pick it up from the environment the python version will match.  E.g. the grobid-home jep was asking for python 3.6 and I have installed 3.8. 
- There is an hardcoded path `lib/python3.8/site-packages/jep` which needs to be fixed, eventually by visiting the CONDA_PREFIX and find the python version. Here we might need to add additional virtual environment support I suppose. 

I can say "It works on my machine" 😂

kermitt2:
Some feedback/question on your changes:

> I've created the task called run-server 

There's already a run task for the service https://github.com/kermitt2/grobid/blob/47cf1a153be64dcec0ffa3245df0a2102f6fc6fa/build.gradle#L328
Isn't redundant?

> and I've extended the library path using the conda environment.

Does it make sense to have JEP native libraries in a python environment - given that it could be conda, python virtual environment, poetry, etc.? It seems opening plenty of cases and compatibility issues (it's also confusing with the DeLFT environment indicated in `grobid.yaml`).

> I've removed the libjep.so from the mac os library path because, depending how it was built might require a python version that is not installed. If we pick it up from the environment the python version will match. E.g. the grobid-home jep was asking for python 3.6 and I have installed 3.8.

This is related to #689. I though first about building automatically with Docker and embedding plenty of JEP native variants, but too complicated...

In the latest version, I added a script to download and install JEP temporary, rebuild locally, and install the platform/python version compatible native JEP library under the GROBID native library path, see https://grobid.readthedocs.io/en/latest/Deep-Learning-models/#classic-python-and-virtualenv (point 4) and ./grobid-home/scripts/install_jep_lib.sh  
I think it's a good way to go with linux and macos platform without too much complexity. 
The script could be called via gradle in the future, and there should then always be a compatible library under grobid-home native library path.

> There is an hardcoded path lib/python3.8/site-packages/jep which needs to be fixed, eventually by visiting the CONDA_PREFIX and find the python version. Here we might need to add additional virtual environment support I suppose.

Following the script, the JEP library would always be under grobid-home, without python specific environment, I think it should work all the time?

lfoppiano:
> In the latest version, I added a script to download and install JEP temporary, rebuild locally, and install the platform/python version compatible native JEP library under the GROBID native library path, see https://grobid.readthedocs.io/en/latest/Deep-Learning-models/#classic-python-and-virtualenv (point 4) and ./grobid-home/scripts/install_jep_lib.sh I think it's a good way to go with linux and macos platform without too much complexity. The script could be called via gradle in the future, and there should then always be a compatible library under grobid-home native library path.

I've checked it out. After more than one week start again it's hard 😅 
I noticed that after checking it out you run `python setup.py build install`. Is the `install` required? I'm asking because the `install` is going to install jep also in local python installation, so I'm not sure is necessary to also copy jep under the grobid-home. 
I've modified the script in local to run on mac os (and removed the `install`) but when I run grobid it lack the correct python version, so I'm wondering, can we really avoid dealing with the mess of the virtual envs?

lfoppiano:
The last commits are mostly commented code, I just want to keep it somewhere 😄

kermitt2:
@lfoppiano  The PR was normally just about setting the native path of JNI JEP under `grobid-home/lib` from gradle, so that we can run Grobid with java 10+ with Linux and Mac OS. We are trying to do several other things here and I think this makes it very confusing and we might never arrive to something that can be merged :) 

1) So I would leave out the part related to the python environment for the moment out of this PR, in particular not mixing the gradle part and the environment. This is very experimental I think. So let's not need to worry here about how the native JEP JNI library is coming under `grobid-home/lib` in this particular PR. 

2) We could do another PR to clarify the JEP install and how to use environment with GROBID+DeLFT. I think we should keep it simple and maintainable. We can simply create an environment for example (with whatever python environment tool), and then do everything with it is activated (install Grobid, run the JEP install script and then start Grobid within this environment). We might not use something else, because a) it is too complicated via Gradle to consider all possible environment cases and paths and b) it currently doesn't work: the python env, path given in the Grobid config is not working and I don't know why (either everything has to run global or everything within a common python environment activated).

3) Setting the JEP JNI library path from Gradle is working with Linux now, but the issue is to have it working also with Mac OS, so make gradle able to detect the platform and set the native path appropriately  `grobid/lib/lin-64`or  `grobid/lib/mac-64`. If we can solve this, we can merge the PR. 

What do you think?

lfoppiano:
@kermitt2 OK for me to just work on assigning the paths for Mac Os and Linux. My bad for insisting on the virtual environment, it's just that I cannot really test without them.

kermitt2:
I went through the difference run cases and adapted gradle for setting `java.library.path` in every cases. 

Tested on Linux-64, it works with JDK 1.15 for the service, training commands, tests and evaluation commands. For the batch mode, the `-Djava.library.path=` need to be added in the command line, which makes it bloated but that's java life - the documentation is updated for that. 

@lfoppiano could you maybe test all these different run cases on Mac ? Normally that should work the same, but just to be sure before merging.

lfoppiano:
I could not install easily JDK 15, so I tested with JDK 17. 
Meanwhile: 
 - I've updated gradle to version 7.2 (7.1.1 did not work due to incompatibility with the JDK bytecode)
 - I've added `jvmArgs "--add-opens", "java.base/java.lang=ALL-UNNAMED"` as parameters to make it working

Issue #923, opened 2022-06-07
pdfalto error: Syntax Warning: Invalid entry in bfchar block in ToUnicode CMap


artturimatias:
Hi,
I'm getting following error with certain pdf:
```
ERROR [2022-06-07 08:02:33,838] org.grobid.core.process.ProcessPdfToXml: pdfalto process finished with error code: 143. [/opt/grobid/grobid-home/pdfalto/lin-64/pdfalto_server, -fullFontName, -noLineNumbers, -noImage, -annotation, -filesLimit, 2000, /opt/grobid/grobid-home/tmp/origin3690432459378499723.pdf, /opt/grobid/grobid-home/tmp/czDhswmAVc.lxml]
ERROR [2022-06-07 08:02:33,838] org.grobid.core.process.ProcessPdfToXml: pdfalto return message: 
Syntax Warning: Invalid entry in bfchar block in ToUnicode CMap
Syntax Warning: Invalid entry in bfchar block in ToUnicode CMap
... LOT of these lines
```

This is the problematic PDF: 
https://jyx.jyu.fi/bitstream/handle/123456789/81469/978-951-39-9321-4_vaitos10062022.pdf?sequence=1&isAllowed=y
Its a dissertation with multiple articles in it.

I'm calling grobid with httpie like this:
`http -f POST :8070/api/processReferences input@'./978-951-39-9321-4_vaitos10062022.pdf;type=application/pdf'`

Same problem also happens via web UI.

OS: Debian 11
Grobid version: 0.7.1 (Docker image)

Any clues what might be causing this?

artturimatias:
I'm not sure how Grobid is calling pdfalto, but I just tried with plain pdfalto: 
`pdfalto file.pdf`
It produces same syntax warnings but it successfully creates xml file(s).

lfoppiano:
@artturimatias which image are you using? there are two `grobid/grobid:0.7.1` and `lfoppiano/grobid:0.7.1`

artturimatias:
Sorry, didn't realise there was two different images. I was using lfoppiano/grobid:0.7.1. I tried grobid/grobid:0.7.1 and it gives same warnings and then time outs.
```

ERROR [2022-06-09 13:11:31,000] org.grobid.service.process.GrobidRestProcessFiles: An unexpected exception occurs. 
! org.grobid.core.exceptions.GrobidException: [TIMEOUT] PDF to XML conversion timed out
[...]

Issue #894, opened 2022-02-25
Problem proccessing PDF: [PDFALTO_CONVERSION_FAILURE]


cerulij:
Hey everyone, I've been having this issue when I try to process a PDF using the Java API. I'm using macOS monterey:

**org.grobid.core.exceptions.GrobidException: [PDFALTO_CONVERSION_FAILURE] PDF to XML conversion failed on pdf file /xxx/Wang_paperAVE2008.pdf** 
	at org.grobid.core.document.DocumentSource.processPdfaltoThreadMode(DocumentSource.java:209)
	at org.grobid.core.document.DocumentSource.pdfalto(DocumentSource.java:155)
	at org.grobid.core.document.DocumentSource.fromPdf(DocumentSource.java:64)
	at org.grobid.core.document.DocumentSource.fromPdf(DocumentSource.java:50)
	at org.grobid.core.document.DocumentSource.fromPdf(DocumentSource.java:42)
	at org.grobid.core.engines.CitationParser.processingReferenceSection(CitationParser.java:396)
	at org.grobid.core.engines.Engine.processReferences(Engine.java:237)

I'm also trying on the pdf in the grobid-example repo (Wang_paperAVE2008.pdf). The code is pretty much the basic usage:

(...)

      Engine engine = GrobidFactory.getInstance().createEngine();

      BiblioItem result = new BiblioItem();

      String tei = engine.processHeader(pdfPath, 0, result);

Do you know what I should do?

Thank you!

NorbertSandor:
I get the same exception on Win10.

kermitt2:
Hello !

@NorbertSandor Windows is not supported anymore by Grobid, the binaries for pdfalto are missing. 

@cerulij it's likely that the binaries for pdfalto shipped with Grobid are not compatible with the recent macOS monterey

@lfoppiano what do you think about the macOS monterey issue?

lfoppiano:
I tested grobid with MacOs Monterey and it works fine. 

@cerulij could you share the whole example or the project so that I could have a look? At first glance might be that the grobid home is not found perhaps?

kermitt2:
@NorbertSandor just pointing to the documentation about the non-support of Windows https://grobid.readthedocs.io/en/latest/Troubleshooting/#windows-related-issues
Using docker is a convenient way to have a Grobid service working on windows, see https://grobid.readthedocs.io/en/latest/Grobid-docker/

karatekaneen:
[...]

PR #926, opened 2022-06-22
Add dynamic timeout on pdfalto_server

lfoppiano:
This PR should add a dynamic timeout on the pdfalto_server script using the argument `--timeout XY`. #923 

It also avoid changing the virtual memory on mac Os. #894


kermitt2:
Thanks a lot Luca !

I tested on Linux 64, but not with Mac Os, and it works fine! 

Note that it breaks compatibility of grobid-home version 0.7.1 when running version 0.7.2-SNAPSHOT, grobid-home will need to be up-to-date too, but there's no other way to make that happen.

PR #932, opened 2022-07-05
Allow absolute paths in temporary directory configuration 

lfoppiano:
This PR allows to use the temporary directory as it is (without concatenation to the grobid-home) when the specified directory is an absolute path.


kermitt2:
Thanks a lot @lfoppiano ! I completely forget about this.

Tested with absolute and relative paths and it's working fine.

kermitt2:
Ok I was a bit too fast, there was an issue for deleting tmp files in the absolute directory.

Issue #933, opened 2022-07-20
Dehypenisation is removing hypens in chemical formulas in certain cases 


lfoppiano:
The input PDF is like this: 
![image](https://user-images.githubusercontent.com/15426/179878282-cad06913-22fe-45e2-aaee-92b05617104b.png)

the data arrives correctly to grobid: 
![image](https://user-images.githubusercontent.com/15426/179878245-72a3c13e-b537-4bfa-b496-e64b639529af.png)

The issue seems to be line 1233 of `TEiFormatter.java`, when calling the dehypenisation: 

```java
String clusterContent = LayoutTokensUtil.normalizeDehyphenizeText(cluster.concatTokens());
```

here the result text:

```[...] low cost rare-earth element) based (Nd 1x Ce x ) 2 Fe 14y Co y B nanostructured magnets. [...]```

I'm working on this issue, I will add details here while I'm moving forward, later I will send a PR. 

Update: 

First issue is here, where we assume that text in the same line has the exact same Y coordinate: 
```
if (breakLine == 0) {
            // check if there is a break-line using coordinates, if not, no dehypenisation
            if (j < tokens.size() && tokens.get(j).getY() == coordinateY) {
                return false;
            }
        }
```
Here the Y coordinate is as follows: 

- `Nd`:  `406.746`
- `1`: `410.506` 
- `-`: `410.506` 
- `x`: `410.538`

Finally it ends at line 284 where the coordinates of `-` and `x` are compared and the Y of `x` is > Y of `-`: 

```
                //I check if the coordinates have changed, this means there is a newline
                if (tokens.get(j).getY() > coordinateY) {
                    return forward;
                }
```

so the tricky part is to estimate correctly what is the amount of Y coordinate that is reasonable for a new line. 

One quick solution could be to do not de-hypenise if the sequence is all subscript or all superscript

lfoppiano:
Another case, in the same paper: 

![image](https://user-images.githubusercontent.com/15426/179888767-ff4dd3f4-5962-4aa3-b996-7576ae7eac60.png)

we have `14-y` at the end of line.

lfoppiano:
Right now the dehypenisation is activated only when a "dash" is encountered. I've add a condition that requires the dash to be neither superscript or subdscript

PR #934, opened 2022-07-20
Avoid dehypenisation when the hypen is superscript or subscript

lfoppiano:
This should resolve issue #933


kermitt2:
Thank you Luca

Issue #160, opened 2017-02-13
Sub/superscript are displayed as plain text characters in the TEI output 


MedKhem:
First re-flexion, identify piece of text as sub/superscript based on position, fonts, etc.

benjaminkreen:
Hey there, I had a quick question. I _just_ started tinkering with grobid and I was wondering if the superscript/subscript identification can be added through training such as giving the following training data:

β-cell Endoplasmic Reticulum Ca<sup>2+</sup>


```xml
<titleStmt>
  <title level="a" type="main">β-cell Endoplasmic Reticulum Ca<sup>2+</sup></title>
</titleStmt>
```

thanks for the input

kermitt2:
subscript and superscript flags are attached to the tokens so we could serialize with `<sup>` and `<sub>` elements yes. 
Similarly we could add `<hi>` for bold and italic tokens.

lfoppiano:
I'm starting to work on implementing this feature. 

What should be done when the token contains combinations? Like `italic + bold`, or `italic+bold+superscript`?

kermitt2:
With the current recognition, the "style" features could support indeed in principle at least italic, bold, superscript/subscript.

The TEI guidelines introduce `<hi>` to encode "graphically distinct" text and there is no constraint on the values, see [here](https://www.tei-c.org/release/doc/tei-p5-doc/en/html/ref-att.global.rendition.html#tei_att.rend). We often see values space-separated (for example `<hi rend="bold italic superscript">`). 

In TEI, there's also `<rendition>` element which uses CSS, which might be more predictable and would not required to further customize the XML schema. 

I think what's complicated are the relations and the possible clash with other structures/tagging. 

* when this style information should be ignored: For instance a reference marker is often in bold or a superscript number. But the logical "reference" structure is already captured by the `<ref>` labeling and there is no point in keeping rendering information here. The `<hi>` is only relevant to text without any other explicit other logical mark-up. The only exception I think would be superscript/subscript inside a formula. 

[...]

PR #936, opened 2022-07-22
Implements font styles in the output XML

lfoppiano:
This PR is implementing the styles italic, bold superscript and subscript in the output xml. 
See information at #160

### DIFF
--- a/grobid-core/src/main/java/org/grobid/core/document/TEIFormatter.java
+++ b/grobid-core/src/main/java/org/grobid/core/document/TEIFormatter.java
@@ -1171,7 +1171,7 @@ public class TEIFormatter {
                 if (CollectionUtils.isNotEmpty(stylesList)) {
                     applyStyleList(head, text, stylesList);
                 } else {
-                    head.appendChild(text);
+                    head.appendChild(StringUtils.normalizeSpace(text.replace("\n", "")));
                 }
 
                 if (config.isGenerateTeiIds()) {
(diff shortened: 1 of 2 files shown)


kermitt2:
Hi @lfoppiano !

This branch will require quite a few tests I think (I suspect it will raise problems to some of the grobid modules and I need to check the consistency with Pub2TEI), so I pushed its release to version 0.8.0. 

One thing related to "document structure" versus "narrative style" is the bold style for section titles. I think it's like the italic/bold for the reference markers, the logical "section title" structure is already captured by the `<head>` element, so I would ignore the style for all section titles. 

For example in the attached pdf, the style should be ignored here:

```xml 
           <div xmlns="http://www.tei-c.org/ns/1.0">
                <head n="1"><hi rend="bold">Introduction</hi></head>
```

In contrast, the style here should be kept because it corresponds to an highlight within the flow of the paragraph text:

```xml
                <p>12. <hi rend="bold">Average tf-idf similarity between citance and title of the cited paper (F12):</hi> We calculate the similarity of each citance with the title of the cited paper and take an average of it.</p>
                <p>13. <hi rend="bold">Maximum tf-idf similarity between citance and title of the cited paper (F13):</hi> We take the maximum of similarity of the citances with the title of the cited paper.</p>
```

Does it make sense?

[qss_a_00170.pdf](https://github.com/kermitt2/grobid/files/11345291/qss_a_00170.pdf)

lfoppiano:
@kermitt2 yes, no problem to push it further. 

OK to the change you propose.

lfoppiano:
The crazy part was to merge the master back in this branch 😅 

> For example in the attached pdf, the style should be ignored here:

I've made the change and now the text within the `<head>` will not have the style applied: 

```
<div
                <head n="1">Introduction</head>
                <p>Literature searches are crucial to discover
```


> In contrast, the style here should be kept because it corresponds to an highlight within the flow of the paragraph text:
```
                <p>12. <hi rend="bold">Average tf-idf similarity between citance and title of the cited paper (F12):</hi> We calculate the similarity of each citance with the title of the cited paper and take an average of it.</p>
                <p>13. <hi rend="bold">Maximum tf-idf similarity between citance and title of the cited paper (F13):</hi> We take the maximum of similarity of the citances with the title of the cited paper.</p>
```

I'm not sure what you mean in this case 🙂

PR #951, opened 2022-09-23
Data and code availability statement zone

kermitt2:
Recognition of data and code availability statements, either in the header or the end of an article and marking the section in a normalized place in the TEI result (similarly as the acknowledgement section). 

This involves the segmentation model (to recognize the zone as an additional section after the main article body) and the header model (when the zone is located within the header). 

Thanks @lfoppiano  :D


kermitt2:
Hi @lfoppiano !

I reviewed everything and made some few changes - in particular put back the header labels into `TaggingLabels.java` to avoid a breaking change in all the current grobid modules (which uses labels from the class TaggingLabels - I actually wanted to put back the segmentation labels too in TaggingLabels to simplify a few years ago, but I didn't do it to avoid a breaking change). 

Currently, when the availability statement is in the header, it is correctly labeled and stored in `BiblioItem.java`, but it is not serialized in the final TEI because the method used can only work with segmentation label:

https://github.com/kermitt2/grobid/blob/feature/data-availability-statement/grobid-core/src/main/java/org/grobid/core/engines/FullTextParser.java#L2473

```java
            // data availability statements
            StringBuilder dataAvailability = new StringBuilder();
            if (StringUtils.isNotBlank(resHeader.getDataAvailability())) {
                dataAvailability = getSectionAsTEI("availability", "\t\t\t", doc, TaggingLabels.HEADER_AVAILABILITY,
                    teiFormatter, resCitations, config);
            } else {
                dataAvailability = getSectionAsTEI("availability", "\t\t\t", doc, SegmentationLabels.AVAILABILITY,
                    teiFormatter, resCitations, config);
            }
```

`getSectionAsTEI()` then uses:

https://github.com/kermitt2/grobid/blob/feature/data-availability-statement/grobid-core/src/main/java/org/grobid/core/engines/FullTextParser.java#L2517

```java
SortedSet<DocumentPiece> sectionPart = doc.getDocumentPart(taggingLabel);
```
Only the `SegmentationLabel` work here with `doc.getDocumentPart()`. 

To retrieve the layout tokens relative to the availability statement from the header stored in `BiblioItem.java`, normally we use:

https://github.com/kermitt2/grobid/blob/feature/data-availability-statement/grobid-core/src/main/java/org/grobid/core/data/BiblioItem.java#L4238

```java
List<LayoutToken> headerAvailabilityStatementTokens = biblio.getLayoutTokens(TaggingLabels.HEADER_AVAILABILITY);
```

kermitt2:
Latest commits fixed the header availability statement TEI serialization.
`createTraining` for header and segmentation now also pre-labeled availability stmt and funding sections.

Issue #957, opened 2022-10-13
Sections belonging to the back of the TEI document are missing from the output when processing only the header 


lfoppiano:
The information that are placed in the back section of the output TEI document (e.g. data availablilty, funding) are currently not in the output when grobid is processing only the header (`processHeader` function). 

This is not a pressing issue because these are not information that users need to have when using the header processing

Issue #652, opened 2020-10-15
Feature Request: Funding Information


de-code:
It would be good to be able to extract funding information.

Unfortunately, the bioRxiv XML itself doesn't contain specific annotations.
i.e. currently they are just mostly back sections.
We would like to extract also things like RRIDs.

Funding related information seem to be mostly contained near the acknowledgements, e.g.:

`183988v1` ([10.1101/183988](https://doi.org/10.1101/183988))

```xml
<ack>
    <title>ACKNOWLEDGEMENTS</title>
    <p>We would like to thank Dr. Maria Bouvy-Liivrand for help with establishing the ST2 cell culture and differentiation and EMBL Gene Core at Heidelberg for support with high-throughput sequencing. The experiments presented in this paper were carried out using the HPC facilities of the University of Luxembourg (<xref ref-type="bibr" rid="c73">Varrette et al., 2014</xref>).</p>
</ack>
<sec id="s5" sec-type="COI-statement">
    <title>COMPETING INTERESTS</title>
    <p>The authors declare they have no competing interests.</p>
</sec>
<sec id="s6">
    <title>FUNDING</title>
    <p>This work was supported by funding from the University of Luxembourg. DG was supported by fellowship from the Luxembourg National Research Fund (FNR) (AFR 7924045).</p>
</sec>
<sec id="s7" sec-type="con">
    <title>AUTHOR CONTRIBUTIONS</title>
    <p>DG, TS and LS conceived the project and designed the experiments and analysis. DG performed all the experiments and DG and LS analyzed the results. DG and AG performed the RNA-seq and ChIP-seq analysis. FS and MHS developed the EPIC-DREM approach. DG, FS and MHS performed the EPIC-DREM analysis. MS performed the Western blot analysis. RH prepared the libraries and performed the sequencing for the AHR-KD experiments. PE developed the randomization method to derive control footprint regions. All authors commented on the manuscript.</p>
</sec>
```

<details>
<summary>GROBID 0.6.1 XML</summary>

```xml
<div type="acknowledgement">
    <div
        xmlns="http://www.tei-c.org/ns/1.0">
        <head>ACKNOWLEDGEMENTS</head>
        <p>We would like to thank Dr. Maria Bouvy-Liivrand for help with establishing the ST2 cell culture and differentiation and EMBL Gene Core at Heidelberg for support with high-throughput sequencing. The experiments presented in this paper were carried out using the HPC facilities of the University of Luxembourg 
[...]

Issue #956, opened 2022-10-13
processShort may yeld loss of text in some cases


lfoppiano:
Working on the funding, I found a small glitch while investigating loss of text. 
In my case, the funding section was empty. 

Example PDF from the PMC set: [main.pdf](https://github.com/kermitt2/grobid/files/9770172/main.pdf)

This is a minor problem, but it could impact several parts: 
![image](https://user-images.githubusercontent.com/15426/195494736-73a549b3-1e5f-41e5-973c-5b7e90828da8.png)

Basically, when we use `processShort()` followed by `TEIFormatter.processTEIDivSection()`: 
 - example on the availability statement https://github.com/kermitt2/grobid/blob/b1c0cfd7c0a92ea85e225d4350d16e89f064379a/grobid-core/src/main/java/org/grobid/core/engines/FullTextParser.java#L2476
 - example on the funding: https://github.com/kermitt2/grobid/blob/b1c0cfd7c0a92ea85e225d4350d16e89f064379a/grobid-core/src/main/java/org/grobid/core/engines/FullTextParser.java#L2506 

The method called within `TEIFormatter.processTEIDivSection()`, in particular `toTEITextPiece()`, around line `TEIFormatter:1159`   expects basically text (not table, not figure) however the processShort could output crazy stuff. 

In the example above the funding statement is tagged as `<table>` for the whole chunk which results in being lost in the final output. 

```
This	this	T	Th	Thi	This	s	is	his	This	BLOCKIN	LINEIN	ALIGNEDLEFT	NEWFONT	HIGHERFONT	0	0	INITCAP	NODIGIT	0	NOPUNCT	0	10	0	NUMBER	0	0	I-<table>
study	study	s	st	stu	stud	y	dy	udy	tudy	BLOCKIN	LINEIN	ALIGNEDLEFT	SAMEFONT	SAMEFONTSIZE	0	0	NOCAPS	NODIGIT	0	NOPUNCT	0	10	0	NUMBER	0	0	<table>
was	was	w	wa	was	was	s	as	was	was	BLOCKIN	LINEIN	ALIGNEDLEFT	SAMEFONT	SAMEFONTSIZE	0	0	NOCAPS	NODIGIT	0	NOPUNCT	0	10	0	NUMBER	0	0	<table>
[...]

PR #959, opened 2022-10-13
Add funding statement in TEI output 

### LINKED ISSUE #895: Paragraph content missing for 'Funding' (annex) section
I am using GROBID to convert bioRxiv preprints PDFs to XML and finding that paragraph content with the heading 'Funding' is not being captured in the TEI XML output (version 0.7.0, lightweight Docker image).

Example PDF:
[2021.09.27.461862v1.full.pdf](https://github.com/kermitt2/grobid/files/8154868/2021.09.27.461862v1.full.pdf)

The 'Funding' section in this document appears immediately after the 'Acknowledgements' on Page 19 of the PDF. It is captured in the TEI XML as `@type="annex"`, but the `<p>` element and text content is missing:

```
<div type="annex">
  <div 
    <head>Funding</head>
   </div>
</div>
```
Although recognition of this section with some `@type="funding"` attribute would be the ideal scenario, for my use case I simply need the associated `<p>` content (i.e. "This work was funded by...") to be output in the TEI XML.

Many thanks in advance for looking into this issue.

lfoppiano:
This PR adds the funding statement (already in training data and models) in the TEI output. 

The funding output will be placed in a standardised position in the `<back>` of the TEI output, as the availability statement. 

e.g. (with sentence segmentation)

```xml
<div type="funding">
<div xmlns="http://www.tei-c.org/ns/1.0"><head>Funding</head><p><s>This study was supported by the International Collaborative Research Grants Scheme with joint grants from the Wellcome Trust UK (GR071587MA) and the Australian NHMRC (268055).</s><s>The funding sources played no role in study design, data collection, analysis or interpretation, writing the report, or the decision to submit the paper for publication.</s></p></div>
</div>
```

This should fix #895 #652 

As discussed in #957, this does not cover the case when only header is processed.


kermitt2:
Thanks ! 
I think the change in [TEIFormatter.java](https://github.com/kermitt2/grobid/pull/959/files#diff-e88bc69454f8143f785f6053c65c9933fec104835298419e9f290a1ea689f44d) is doing nothing? Just moving stuff that do the same (simply put a default space at inline position of a figure to avoid agglutination of text).

kermitt2:
From the discussion https://github.com/kermitt2/grobid/issues/956, I am updating the PR with post-processing limited to processShort(), instead  of changing the general fulltext model decoding.

kermitt2:
The changes seem to work fine, on the example from #956 we obtain:

```xml
           <div type="funding">
                <div
                    <p>This study was supported by the South Asian Clinical Toxicology Research Collaboration, which is funded by The Wellcome Trust/ National Health and Medical Research Council International Collaborative Research Grant GR071669MA. The funding bodies had no role in analyzing or interpreting the data or writing the article.</p>
                </div>
            </div>
```
No runtime error on PMC_sample set and biorxiv_test_2000.

lfoppiano:
I found several cases where where the papers has the funding statement in the back of the document, the TEI serialisation calls `getSectionAsTEI()` which does not uses the `processShort()`.

E.g. this document: [bj4360053.pdf](https://github.com/kermitt2/grobid/files/9797087/bj4360053.pdf)

Unfortunately, the result does not help, and the hack in processShort() does not work because of the different label at the first place won't fix the sequence. 

```
FUNDING	funding	F	FU	FUN	FUND	G	NG	ING	DING	BLOCKSTART	LINESTART	ALIGNEDLEFT	NEWFONT	HIGHERFONT	1	0	ALLCAP	NODIGIT	0	NOPUNCT	0	10	0	NUMBER	0	0	I-<section>
This	this	T	Th	Thi	This	s	is	his	This	BLOCKSTART	LINESTART	ALIGNEDLEFT	NEWFONT	LOWERFONT	0	0	INITCAP	NODIGIT	0	NOPUNCT	0	10	0	NUMBER	0	0	I-<table>
work	work	w	wo	wor	work	k	rk	ork	work	BLOCKIN	LINEIN	ALIGNEDLEFT	SAMEFONT	SAMEFONTSIZE	0	0	NOCAPS	NODIGIT	0	NOPUNCT	0	10	0	NUMBER	0	0	<table>
was	was	w	wa	was	was	s	as	was	was	BLOCKIN	LINEIN	ALIGNEDLEFT	SAMEFONT	SAMEFONTSIZE	0	0	NOCAPS	NODIGIT	0	NOPUNCT	0	10	0	NUMBER	0	0	<table>
supported	supported	s	su	sup	supp	d	ed	ted	rted	BLOCKIN	LINEIN	ALIGNEDLEFT	SAMEFONT	SAMEFONTSIZE	0	0	NOCAPS	NODIGIT	0	NOPUNCT	0	10	0	NUMBER	0	0	<table>
[...]
```

I'm not sure how I could fix this properly

kermitt2:
We could wait for future versions which will remove the tables and figures from the fulltext model, or apply the same fix as in processShort for every table and figure labels (not just the ones starting the sequence). What do you think?

kermitt2:
So the two problematic cases are working fine with the current version, yeah :D 

I cleaned the previous hack in `processShort()` and replaced it with `postProcessFullTextLabeledText()`, to be sure it is always applied after calling `processShort()` (so this applied to the funding statement in header too now, otherwise it was only applied to the funding statement at the end of a paper).

kermitt2:
Thanks !!

PR #978, opened 2022-12-06
add some training data with examples with docs containing highlights

Aazhar:
some documents contains highlights in the first page, this part is confused with the abstract, I'm submitting some training data for this case


lfoppiano:
HI @achrafazharccsd apologize for checking this 2 years in late. 

There is a problem with these articles, both of them are not CC-BY. The Cell's one is CC-BY-ND-NC (ND = Non derivative), which does not allow derivative work. The `Pacman` is not CC. So both of them cannot be really added to the grobid public data. 

Also the annotation require some correction (I checked them before realising the licence), If you have some equivalent articles that are CC-BY, I'm happy to help you with the training data and providing an updated model for it.

Issue #862, opened 2021-11-18
error 500


shainaraza:
I install even verison 0.7.0 but gets this error

**Error encountered while requesting the server.
Response 500: - The PDF document cannot be annotated. Please check the server logs.**

any help

kermitt2:
Hello @shainaraza !

Server log are under grobid/logs/grobid-service.log

You should find the issue in the logs, or you could attach it here so that we can have a look.

shainaraza:
[](url)is it version conflict, I just using openjdk 8.
here is log
[grobid-service.log](https://github.com/kermitt2/grobid/files/7565244/grobid-service.log)

shainaraza:
even i use the example file from it, gets the same error

kermitt2:
Thanks for the log file. 

Windows is not supported any more by Grobid 0.7.0, pdfalto has not been recompiled for Windows 64 (lack of time/skills) and according to the log this is the problem.  

You could use the docker image for using Grobid on Windows?

shainaraza:
no I dont have a docker image, any alternative?

shainaraza:
i try using COlab environment but it does not allow localhost

kermitt2:
If you can't install docker on Windows, I think some people managed to get Grobid working the Windows subsystem/shell for Linux, but I am very ignorant on the matter.

shainaraza:
hi @kermitt2 Now i am on linux system with all setup but I still get the same error 500, please see log
[grobid-service.log](https://github.com/kermitt2/grobid/files/7572162/grobid-service.log)

kermitt2:
According to the logs, the PDF is empty or it is a PDF image only:

```
! org.grobid.core.exceptions.GrobidException: [NO_BLOCKS] PDF parsing resulted in empty content
```

Do you have a text layer selectable on this PDF ?

shainaraza:
@kermitt2 thanks you for replying me. I dont know text layer selection?
in my case, both in Windows and Linux now, the localhost:8070 works and let me upload a file but gets 500 ERROR after that.
advise please

kermitt2:
If you open the PDF in a PDF viewer, can you select the text? 
You can also try the `pdftotext` command line on Linux, if it returns some text. 
If there is no text, a preliminary OCR is required.

shainaraza:
yes I can definitely select the text.
I also use TIKA and pdftotext , the reason I want to use Grobid is because of its structure, metadata that it returns.
[...]

PR #1007, opened 2023-04-30
Temporary fix to Windows pdfalto issue #862

AaronNGray:
Have modified the generated command on Windows to work without the `-noLineNumbers` and the `--timeout XXX` options.

NOTE: This does not fix all issues on Grobid on Windows !


AaronNGray:
aborted due to incomplete commit :(

PR #1010, opened 2023-05-06
Support for python env managers 

lfoppiano:
This PR adds support for python virtual environments for JDK 11+. Should be merged after #921. 
It supports venv and conda. 

Update: I've noticed I'm not able to run grobid natively on linux without installing the python dependencies as a system dependencies. Usually I've correlate the use of this PR with the removal of the JEP library under the grobid-home, in favour of the JEP installed in the python library / virtual environment. 

We could improve the detection of the python version.


lfoppiano:
With this PR is possible to run a different version than 3.8 (the jep shipped with grobid requires python 3.8) and relay on conda for the environment. 

Here what I did for running DL models natively using a conda environment: 
1. cd grobid 
1. Switch to this branch 
2. `rm -rf grobid-home/lib/lin-64/jep` 
3. `cd ..`
4. `git clone https://github.com/kermitt2/delft` (delft should be in the parent directory)
5. assuming conda is installed and the environment `delft` has been created with either python=3.8 or python=3.9 (e.g. `conda create --name delft python=3.8`
6. `pip install delft==0.3.4`
7. `conda install jep=4.1.1`  (conda will install also openjdk 8)
8. `conda uninstall --force openjdk-8`
9. `conda install --force openjdk-17`
10. change any model in the grobid config to use delft instead of wapiti (e.g. header model)
11. ./gradlew run

PR #1015, opened 2023-05-17
Enable Github actions / Disable circleci 

lfoppiano:
This PR configures correctly github actions to build, test and send the tests statistics to coveralls.io 

Since it was trial and error I had to push several time. I recommend to merge and squash it in a single commit, to keep the history clean.


kermitt2:
Thank you Luca ! I think I am too lazy to try to squash it in a single commit :D

PR #1017, opened 2023-05-22
Completing translation

ACMCMC:


ACMCMC:
Oh, got it!! I just didn't understand it when I read it so I assumed it was a mistake haha :) Happy to know it was on purpose!!

kermitt2:
I suppose only German people or people having lived in Germany (it was my case at the time) would understand it, well probably not a good idea :D 
Thanks !

Issue #1019, opened 2023-05-25
libpthread.so.0: undefined symbol: __libc_pthread_init, version GLIBC_PRIVATE


jimstein3d:
I am using Ubuntu 22.04.2 LTS both in WSL2 and as VM on Hyper-V with Java version 17

>jimmy@LAPTOP-JIMMYE:~/grobid$ java --version
openjdk 17.0.7 2023-04-18
OpenJDK Runtime Environment (build 17.0.7+7-Ubuntu-0ubuntu122.04.2)
OpenJDK 64-Bit Server VM (build 17.0.7+7-Ubuntu-0ubuntu122.04.2, mixed mode, sharing)

Build succeeds - but get error when running `:grobid-service:run`,`./gradlew clean install test` or `createTraining`.

What went wrong:
>/usr/lib/jvm/java-17-openjdk-amd64/bin/java: symbol lookup error: /home/jimmy/grobid/grobid-home/lib/lin-64/libpthread.so.0: undefined symbol: __libc_pthread_init, version GLIBC_PRIVATE

I do not know what to do about that some dependencies reference my system libraries.  Is it important that some listed dependent libraries dont have a local path (to grobid/grobid-home/lib/lin-64/)?

>jimmy@LAPTOP-JIMMYE:~$ ldd grobid/grobid-home/lib/lin-64/libwapiti.so
        linux-vdso.so.1 (0x00007ffea9591000)
        libstdc++.so.6 => /home/jimmy/grobid/grobid-home/lib/lin-64/libstdc++.so.6 (0x00007fba68c04000)
        libm.so.6 => /home/jimmy/grobid/grobid-home/lib/lin-64/libm.so.6 (0x00007fba68ab5000)
        libgcc_s.so.1 => /home/jimmy/grobid/grobid-home/lib/lin-64/libgcc_s.so.1 (0x00007fba68a9a000)
        libpthread.so.0 => /home/jimmy/grobid/grobid-home/lib/lin-64/libpthread.so.0 (0x00007fba68a77000)
        libc.so.6 => /home/jimmy/grobid/grobid-home/lib/lin-64/libc.so.6 (0x00007fba68885000)
        /lib64/ld-linux-x86-64.so.2 (0x00007fba68e07000)

I have already tried to `git checkout tags/0.7.3`

Stacktrace:
jimmy@LAPTOP-JIMMYE:~/grobid$ ./gradlew run --stacktrace

> Task :grobid-service:run FAILED
[...]

PR #1020, opened 2023-05-29
Rebuild with older glibc version for better portability

kermitt2:
Fix for #1019 - rebuild libwapiti.so with glibc 2.17 to rely on backward compatibility, so remove other libs used before for local linking.


kermitt2:
Tested with Ubuntu 20.04 and 22.04.

Issue #1027, opened 2023-06-07
On HuggingFace demo, content type for processHeader is bibtex, while default should be XML


kermitt2:
On the current HF demo at https://kermitt2-grobid.hf.space (based on 0.7.3 full docker image), processHeader returns bibtex

![Screenshot from 2023-06-07 17-21-40](https://github.com/kermitt2/grobid/assets/2340795/45d9a6ce-b1d2-4c82-9faa-583962a05e1e)

While the normal default should be XML, below current master on local with JDK 8 

![Screenshot from 2023-06-07 17-23-13](https://github.com/kermitt2/grobid/assets/2340795/5d621cc6-0ce8-4ea0-be94-3df0154b5441)

Issue #1023, opened 2023-06-06
jQuery < v3.5.0 contains Cross-site vulnerabilities - how to fix in Docker and non-docker deployments?


mfeblowitz:
- OS: Docker on ubuntu

- Java version: openjdk 17.0.2 2022-01-18, as installed in docker container

- Problem: multiple cross-site scripting vulnerabilities exist in both Docker and non-docker ground service deployments (https://www.tenable.com/plugins/nessus/136929). This should be fixed for all Grobid deployments. 

For non-docker installations, a simple replacement of the current references to older jQuery versions is possible:
```
sed -i 's/jquery-3.4.1.min.js/jquery-3.5.1.min.js/' ./grobid-service/src/main/resources/web/index.html
rm ./grobid-service/src/main/resources/web/resources/js//jquery-3.4.1.min.js
wget  https://code.jquery.com/jquery-3.5.1.min.js --no-check-certificate
mv jquery-3.5.1.min.js ./grobid-service/src/main/resources/web/resources/js/
```
How can this be done for the ground docker image?

kermitt2:
Sorry for the damn auto-close!

PR #1028, opened 2023-06-08
Update jquery to version > 3.5.0 to avoid XSS + bibtex fix

lfoppiano:
This should solve #1023 and #1027


kermitt2:
Thanks a lot ! It should fix https://github.com/kermitt2/grobid/issues/1027 yes, we'll see for the next docker build.

lfoppiano:
I've created a new branch 0.7.3-fixes and cherry-picked these changes there. 
I think we can build the 0.7.3 docker image again and replace the old one.

kermitt2:
> In the HF space, in the embedded version the bibtex is still returned.. There is something related to the HF stuff that is changing the request from xml to text I think.

With slightly older docker image, it was working, for example https://huggingface.co/spaces/kermitt2/grobid-crf running with 0.7.3.-SNAPSHOT is correctly managing XML for processHeader. I was rather suspecting something with the JDK version (more or less the only thing that changed between the images), but weird in any cases.  

Screenshot from 2023-06-13 08-42-41

lfoppiano:
Sorry, good news.. it's working.. It seems that I did not properly refresh.. 

image

kermitt2:
Cool ! I'll update the full image too.

kermitt2:
It's done for the full image on Docker Hub too, and it works fine on HF.

PR #1029, opened 2023-06-08
Minor update to the docker images

lfoppiano:
Remove the mac arm libraries from the linux-based docker images. 
Until we build a multi-arch image we don't need these libraries to be there.


kermitt2:
Thanks !

PR #1030, opened 2023-06-08
avoid 500 when the equations are null

lfoppiano:
Found this while testing #1028, it happened when the getEquations() is null. It happened with this PDF 😅 
[Curriculum_Vitae.pdf](https://github.com/kermitt2/grobid/files/11683013/Curriculum_Vitae.pdf)


kermitt2:
Hi Luca ! As a note I could not reproduce the 500 on the document (great CV, super interesting projects :D ) with the current master. But better to add checks in any cases.

lfoppiano:
😂

PR #1069, opened 2023-12-18
Update affiliation process

kermitt2:
This PR updates the usage of the affiliation-address model:
- review of features
- a bit of training data correction
- update the models
- using clusteror for decoding the labeled result (much simpler and clean)
- optional PDF coordinates for affiliation structure in the resulting TEI XML

For best results, use  the Deep Learning model, which gives significantly more accurate recognition.


kermitt2:
Still to do: 
- update documentation for affiliation coordinates,
- consider coordinates for sub-structures of the affiliation orgName and address elements (country, etc.)
- tests

PR #1070, opened 2023-12-20
correct <title> coordinates attribute name

lfoppiano:
The attribute for the coordinates in titles is 'coord' instead of 'coords'. 

I assume it was a typo, and it should be consistent with the rest of the TEI output, anyway it was easier to submit a PR than to ask first :smile:


lfoppiano:
I also noticed that the title coordinates are not showing in the web interface, so I added it to the list in the Javascript file. 

I also found that not all elements are present here, although this is for batch processing I suppose: 

https://github.com/kermitt2/grobid/blob/6bd974d4c4f0db99904525ba12dc064c10052536/grobid-core/src/main/java/org/grobid/core/engines/ProcessEngine.java#L140

kermitt2:
Yes typo ! I found a couple others for <title> element :/

Also removed the `p` in grobid.js because it is for another PR.

kermitt2:
Thank you for the fix Luca !

PR #1088, opened 2024-02-28
Reduce docker image size for the NN grobid version 

lfoppiano:
This PR attempts to reduce the docker image by reducing the number of layers. 

By just chancing the layers, I managed to save 700Mb. We could optimise more in future, especially considering that grobid modules (e.g. grobid-quantities) are not able to build properly at the moment, on github actions (limited to ~20Gb). 

One solution would be to move everything in the build and then just copy the resulting files in the tensorflow image. 

I also noticed that the delft used here is 0.3.3, I'm not sure whether it's by design, I propose to update it to 0.3.4.

### DIFF
--- a/Dockerfile.delft
+++ b/Dockerfile.delft
@@ -67,12 +67,12 @@ FROM tensorflow/tensorflow:2.7.0-gpu
 ENV LANG C.UTF-8
 
 # update NVIDIA Cuda key (following a key rotation in April 2022)
-RUN apt-get install -y wget
-RUN apt-key del 7fa2af80
-RUN rm /etc/apt/sources.list.d/cuda.list
-RUN rm /etc/apt/sources.list.d/nvidia-ml.list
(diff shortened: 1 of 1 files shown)


kermitt2:
Thank you very much Luca, this was indeed requiring some improvements !
The changes in DeLFT 0.3.4 did not impact Grobid, but good to update anyway.

lfoppiano:
@kermitt2 FYI I've created a new temporary image: `lfoppiano/grobid:0.8.0-full-slim` that I need to be able to build grobid-quantities through github actions. 

As a reference the effficiency climbed from 88% to 91% according to https://github.com/wagoodman/dive

lfoppiano:
Two more noted: 
1. delft 0.3.4 is in pipy but not tagged in the repository
2. it seems that blingfire has been added in the reader.py, however is not specified in the requirements.txt so any grobid docker image are failing when running using delft

PR #1094, opened 2024-04-10
Add additional github actions

lfoppiano:
This PR will extend the current github actions: 
- The automatic build only produce the CRF-only image (`latest-develop`)
- We added a manual build that will push both CRF-only (tags `latest-develop`, `latest-crf`) and the delft full image (tag `latest-full`) . The user can select any branch, this is useful for testing specific features.


lfoppiano:
As discussed, I've removed the docker build of the full image from the automatic build (which will build only the CRF image). The full image could be still built manually from the github actions.

PR #1097, opened 2024-04-12
Avoid splitting URLs between sentences

lfoppiano:
This PR addresses the issue of the sentence segmenter that might split URLs between sentences. 
Updating the regex `urlPattern` is hard to do without high risk of introuducing new bugs (some experiments/attempts [here](https://regex101.com/r/vGo0lM/1))

The original grobid method to exploit the URI pdf annotations, was extended to support cases where the layout token resulting text and the provided postprocessed text differs which was leading to OutOfBoundException. 

We have added/modified the following methods: 

- new method `public static List<OffsetPosition> characterPositionsUrlPatternWithPdfAnnotations(List<LayoutToken> layoutTokens, List<PDFAnnotation> pdfAnnotations)` returns the character offset position **in respect of the layout token string** (that could be obtained by LayoutTokenUtil.toText(tokens). 
- new method `public static List<OffsetPosition> tokenPositionsUrlPatternWithPdfAnnotations(List<LayoutToken> layoutTokens, List<PDFAnnotation> pdfAnnotations)` returns the token offset position.
- modified the original `public static List<OffsetPosition> characterPositionsUrlPatternWithPdfAnnotations(List<LayoutToken> layoutTokens, List<PDFAnnotation> pdfAnnotations, String text)` returns the character offset position **in respect of the text string that is passed in input**. 
 
~~There are often cases where the `text` string and the aggregated string from the layoutToken are not matching (e.g. the text string is dehypenised), and this causes OutOfBoundException when applying substring.~~

The last method (`characterPositionsUrlPatternWithPdfAnnotations(List<LayoutToken> layoutTokens, List<PDFAnnotation> pdfAnnotations, String text)`) is called when the sentence segmenter is running so that we avoid splitting sentences with a URL in between. 

The PR #1099 will improve the recognition because, in this PR, by applying the fix in the sentenceSegmenter that takes text as a string, the process is applied to the layout tokens and not to the text that might be dehypenised, and desynchronised with the layout tokens.


lfoppiano:
This issue was tested by processing all PMC and Biorxiv documents. No error or failures during processing. 
Then I inspected the URLs with regexes and verified that no URLs were over sentences.  

I also tested a bunch of problematic PDF documents.

PR #1099, opened 2024-04-15
Identify URLs and output them in TEI

lfoppiano:
This PR is a continuation of #1097. In this PR the URLs are output in the XML-TEI under the tags `<ref type="url" target="cleaned_version">url appearing from the data extraction</ref>`. 

As with #1097 we exploit the PDF annotations for clickable URLs to correct the extracted information from the regular expressions ([I tried to improve them](https://regex101.com/r/vGo0lM/1) but did not obtained anything more robust). 

The extracted URLs are aggregated to the notes and processed in the same data flow, thus with different formatting rules, as they are appearing in the text. This change will help in future if we need to add the identification of more elements. The implementation could be improved with more specialised data structures instead of the `Triple` object, to carry out the information. 

Here some output examples: 

```xml
<div type="availability">
    <div
        <head>Data availability statements</head>
        <p>
            <s>Google Earth Engine applications to visualize the datasets:
                <ref type="url" target="https://github.com/shijuanchen/shift_cult">
                    https://github.com/shijuanchen/shift_cult
                </ref>
                Map products visualization:
                <ref type="url" target="https://sites.google.com/view/shijuanchen/research/shift_cult">
                    https://sites.google. com/view/shijuanchen/research/shift_cult
                </ref>
            </s>
        </p>
        <p>
            <s>The data that support the findings of this study are openly available at the following URL/DOI:
                <ref type="url" target="https://doi.org/10.5281/zenodo.7782782">https://
                    doi.org/10.5281/zenodo.7782782</ref>.
            </s>
        </p>
    </div>
</div>
```

Here an example with sentence segmentation: 
```xml
<p>
    <s>We compared our area estimates of New Shifting Cultivation with the official forest change statistics from Laos (table
        <ref type="table" target="#tab_1">S1</ref>).
    </s>
    <s>The Laos official forest change maps (
        <ref type="url" target="https://nfms.maf.gov.la/">https://nfms.maf.gov.la/</ref>) are created from the land cover classification maps from the start year and end year for each period (see the periods in table
        <ref type="table" target="#tab_1">S1</ref>).
    </s>
    <s>Since shifting cultivation is the major driver of forest degradation and deforestation in Laos, we expect that there are some consistencies between the areas of New Shifting cultivation and the areas of forest degradation and deforestation.</s>
    <s>There are consistencies in the period
        <ref type="bibr">2006-2010 and 2011-2015, with</ref> the differences between our estimates and the official statistics both less than 1% of Laos.
    </s>
    <s>Our estimates of New Shifting Cultivation are generally higher than the Laos official estimates of deforestation and forest degradation, except for 2006-2010.</s>
    <s>This was partly due to the different monitoring approaches.</s>
    <s>Without using dense time series, the shifting cultivation events that occurred over five years may be difficult to detect using two classification maps from the start and the end.</s>
    <s>In the period
        <ref type="bibr">2001-2005 and 2016-2020, our</ref> estimates are about 2%-3% higher than the official estimates.
    </s>
    <s>For 2016-2020, the discrepancy is partly because the 2019 and 2020 changes are included in our estimates but not in the official statistics.</s>
    <s>Overall, our results and area estimates provide valuable information regarding the forest dynamics of Laos.</s>
</p>
```

```xml
<p>This work is available at
    <ref type="url" target="https://github.com/lfoppiano/supercon2">https://github.com/lfoppiano/ supercon2</ref>. The
    repository contains the code of the SuperCon 2 interface, the curation workflow, and the Table
    <ref type="table">2</ref>. Data support, the number of entities for each label in each of the datasets used for
    evaluating the ML models. The base dataset is the original dataset described in
    <ref type="bibr" target="#b17">[18]</ref>, and the curation dataset is automatically collected based on the database
    corrections by the interface and manually corrected.
</p>
```

with sentence segmentation 
```xml
<div
    <head n="5.">Code availability</head>
    <p>
        <s>This work is available at
            <ref type="url" target="https://github.com/lfoppiano/supercon2">https://github.com/lfoppiano/
                supercon2</ref>.
        </s>
        <s>The repository contains the code of the SuperCon 2 interface, the curation workflow, and the Table
            <ref type="table">2</ref>. Data support, the number of entities for each label in each of the datasets used
            for evaluating the ML models.
        </s>
        <s>The base dataset is the original dataset described in
            <ref type="bibr" target="#b17">[18]</ref>, and the curation dataset is automatically collected based on the
            database corrections by the interface and manually corrected.
        </s>
    </p>
</div>
```

TODO: 
~~- fix some corner cases, where the `.` or the `)` are wrongly retained, these are cases when the regex catches too much (usually one or a few characters)~~
~~- sometimes the URL refs are attached to the following text, it seems that a space is missing somehow~~
~~- validate the XML-TEI schema~~

### DIFF
--- a/.github/ISSUE_TEMPLATE/general-report.md
+++ b/.github/ISSUE_TEMPLATE/general-report.md
@@ -4,11 +4,12 @@ about: Create a report to help us improve
 title: ''
 labels: ''
 assignees: ''
-
 ---
 
-- What is your OS and architecture? Windows is not supported and Mac OS arm64 is not yet supported. For non-supported OS, you can use Docker (https://grobid.readthedocs.io/en/latest/Grobid-docker/)
+Before creating a new issue, make sure you had a look at the [official documentation](https://grobid.readthedocs.com). For specific questions, you can try the [Mendable Q/A chat](https://www.mendable.ai/demo/723cfc12-fdd6-4631-9a9e-21b80241131b) (**NOTE**: This is rather experimental, if not sure, make sure you double-check using the official documentation.)
+
+- What is your OS and architecture? Windows is not supported and Mac OS arm64 is experimentally supported. For non-supported OS, you can use Docker (https://grobid.readthedocs.io/en/latest/Grobid-docker/)
 
 - What is your Java version (`java --version`)? 
 
-- In case of build or run errors, please submit the error while running gradlew with ``--stacktrace`` and ``--info`` for better log traces (e.g. `./gradlew run --stacktrace --info`) or attach the log file `logs/grobid-service.log`.
+- In case of build or run errors, please submit the error while running gradlew with ``--stacktrace`` and ``--info`` for better log traces (e.g. `./gradlew run --stacktrace --info`) or attach the log file `logs/grobid-service.log` or the console log. 
--- a/.github/workflows/ci-build-manual-crf.yml
+++ b/.github/workflows/ci-build-manual-crf.yml
@@ -1,0 +1,42 @@
+name: Build and push a CRF-only docker image
+
+on: 
+  workflow_dispatch: 
+    
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v4
+    - name: Set up JDK 17
+      uses: actions/setup-java@v4
+      with:
+        java-version: '17.0.10+7'
+        distribution: 'temurin'
+        cache: 'gradle'
+    - name: Build with Gradle
+      run: ./gradlew build -x test
+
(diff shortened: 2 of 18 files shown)


lfoppiano:
I've solved additional corner cases: 
1. there are clickable links but no annotation are extracted by grobid, sometimes the regex alone catches too much, e.g. end parenthesis. This was fixed 
2. Spaces before the annotation are only added in the case of URLs, if there is a trail space in the layout token (but not in the dehypenised text)
3. Some other case where the PDF document annotation are wrong (e.g. cut over the breaking line, or incorrect, the resulting URL might be incorrect

Issue #1067, opened 2023-12-05
Questions about data annotation in GROBID


SirLYC:
Hello kermitt2,

I'm currently working on annotating data for the GROBID project and have a few questions regarding the annotation process. I would appreciate it if you could provide some guidance on the following issues:

1. In the `General Principles` section of the documentation, it is mentioned that the text flow should not be changed. Does this mean that the order and content of the text flow in the pre-annotated data cannot be altered or removed? Or does it mean that the internal content of each XML text node cannot be modified, but the external order can be freely adjusted? For example, in the pre-trained data of `references.referenceSegmenter.tei.xml`, due to the PDF text editing order issue, the extracted text flow contains the content after each reference's number first, followed by the number's content, and some main text content mixed in between. In this case, am I allowed to:
    - Remove non-reference content
    - Move the text flow of the reference number to its corresponding reference entry

2. When I finish modifying `segmentation.xml` and proceed to modify `fulltext.xml`, I find that some content inputted into `fulltext.xml` does not belong to the body, or some body content is recognized as front during the segmentation stage. In this case, should I remove the content that does not belong to the body and add back the missing content in the body? Additionally, I would like to know if I am allowed to adjust the order of text tokens if the extracted body order does not conform to the human reading order (while still ensuring that the text child nodes remain unchanged)?

I look forward to your response, and thank you for your assistance!

Best regards,
[Your Name]

kimn1944:
I would like to second @SirLYC's point num 2. In my own testing I see that a piece of text marked as <body> in the segmentation.xml is incorrectly showing up in the header.xml instead of fulltext.xml. I also wonder if we are able to retrain this behavior by moving that element from header.xml to fulltext.xml.

lfoppiano:
Dear @SirLYC 
   sorry for the late answers. 

1. The text flow should not be modified, which means that you can only add, move or remove the HTML tags that define each entity. If the data is out of order because of the transformation from PDF, it should not corrected otherwise the model will learn a condition that will not likely happen. 

2. In general I prefer to work transversely because certain annotations might be complex and we become more efficient in correcting them by type rather than by document. So I would first correct all segmentation files (ignoring the ones that are already corrected), then when finished, re-train the segmentation model, re-generate the training data, and move to the next model in the cascade: e.g. full text or header.

PR #1105, opened 2024-04-25
Fix heading annotation in fulltext evaluation and add header levels

Schroedi:
`<head>Methods<lb/> Preparation<lb/></head>` are actually two headings in the original paper.

### DIFF
--- a/grobid-trainer/resources/dataset/fulltext/evaluation/tei/PMC2644281.training.fulltext.tei.xml
+++ b/grobid-trainer/resources/dataset/fulltext/evaluation/tei/PMC2644281.training.fulltext.tei.xml
@@ -4,11 +4,12 @@
 		<fileDesc xml:id="0"/>
 	</teiHeader>
 	<text xml:lang="en">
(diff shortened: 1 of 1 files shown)


lfoppiano:
@Schroedi Great.

Did you check also the segmentation model training output files? If they were perfectly fine we should not include them, but if they had to be corrected, then we should also add them. 

See comment here: https://github.com/kermitt2/grobid/issues/1067#issuecomment-1888503015

Schroedi:
The segmentation looked fine to me.

Could you share a folder with the PDFs the (fulltext) training data are from? I am currently fetching them one by one.

kermitt2:
hi @Schroedi please send me an email and I will send you back the info for accessing the PDF repository used for the training data.

lfoppiano:
@Schroedi I'm going through the PRs and I was wondering if this PR is completed. I see only one file changed for this PR however I see there are other files for which the `level` attribute can be added. 

I also wantes to thank you anyway for your contributions and sorry for being slow in integrating into Grobid.

Issue #1090, opened 2024-03-08
Funding, acknowledgement statements are not split into sentences 


lfoppiano:
I've noticed that while the data availability is split into sentences, the funding statement is not. Is this by design or should be implemented? 

Example: 
```xml
		</body>
		<back>

			<div type="funding">
<div xml:id="_ERHBmGS"><p xml:id="_WYZCd2J">Funding: This work was supported by the <rs type="funder">National Natural Science Foundation of China</rs> (<rs type="grantNumber">51561009</rs>), the <rs type="funder">Natural Science Foundation of Jiangxi Province</rs> (<rs type="grantNumber">20192BAB206004</rs> and <rs type="grantNumber">20202BAB214003</rs>), the <rs type="funder">Key Research and Development Program of Jiangxi Province</rs> (<rs type="grantNumber">20202BBE53014</rs>), the <rs type="funder">Open Foundation of Guo Rui Scientific Innovation Rare Earth Functional Materials Co</rs>., Ltd.(<rs type="grantNumber">KFJJ-2019-0004</rs>), the <rs type="funder">Doctoral Start-up Foundation of Jiangxi University of Science and Technology (205200100110)</rs>, and the <rs type="funder">Foundation of Jiangxi Educational Department</rs> (<rs type="grantNumber">GJJ200832</rs> and <rs type="grantNumber">GJJ190478</rs>).Institutional Review Board Statement: Not applicable.Informed Consent Statement: Not applicable.</p></div>
			</div>
			<listOrg type="funding">
				<org type="funding" xml:id="_3wHAxav">
					<idno type="grant-number">51561009</idno>
				</org>
				<org type="funding" xml:id="_NRNDwrU">
					<idno type="grant-number">20192BAB206004</idno>
				</org>
				<org type="funding" xml:id="_uNWJMnb">
					<idno type="grant-number">20202BAB214003</idno>
				</org>
				<org type="funding" xml:id="_2MPuZAy">
					<idno type="grant-number">20202BBE53014</idno>
				</org>
				<org type="funding" xml:id="_B7kBgef">
					<idno type="grant-number">KFJJ-2019-0004</idno>
				</org>
				<org type="funding" xml:id="_tk7RJ29">
					<idno type="grant-number">GJJ200832</idno>
				</org>
				<org type="funding" xml:id="_mCAyMcx">
					<idno type="grant-number">GJJ190478</idno>
				</org>
			</listOrg>

			<div type="availability">
[...]

PR #1106, opened 2024-04-28
Add missing sentence segmentation in funding and acknowledgement 

lfoppiano:
This PR implement fixes the way the `funding-acknowledgment` parser handles an already formatted statement and preserve their existing elements (references, sentences). #1090

This PR fixes the following problems: 
 - sentence segmentation lost for funding and acknowlegment statements
 - reference markers are lost after the funding-acknowlegment parser is applied 

The initial solution proposed in the #1090 discussion, to re-apply a sentence segmentation that act on the transformed TEI structure was not applicable because it was not possible to re-generate the sentence coordinates as the TEI-XML elements do not have anymore layout-token information. 

NOTE: this should be merged after #1096


lfoppiano:
I've started testing and noticed that in rare cases (although possible), the sentence segmentation, which is performed before the funding-acknowledgment model, result in sentences that fall on funding-acknowledgment annotations. 

e.g. The first is the original version, without the sentence segmentation: 

```xml 
<div type="acknowledgement">
    <div>
        <head coords="31,72.00,491.09,114.40,12.58">Acknowledgments:</head>
        <p coords="31,191.82,493.44,347.12,9.57;31,72.00,522.72,81.26,9.57">We thank
            <rs type="person">Drs. Carsten Korth</rs> and
            <rs type="person">Nick Brandon</rs> for generously providing anti- DISC1 antibodies.
        </p>
    </div>
</div>
```

Here the first sentence falls on the annotation "Drs.Carsten Korth":

```xml
<div type="acknowledgement">
    <div>
        <head>Acknowledgments:</head>
        <p>
            <s>We thank Drs.</s>
            <s>Carsten Korth and
                <rs type="person">Nick Brandon</rs> for generously providing anti- DISC1 antibodies.
            </s>
        </p>
    </div>
</div>
``` 

I've then worked out a solution that allow merging and updating sentences that are in this situation, including their coordinates.

lfoppiano:
I did run all documents from PLOS, PMC and biorxiv over this, I've checked manually differences for a sample from each corpus. I also checked manually the merging of the coordinates for certain problematic documents that were causing problems in the past. It seems that everyhting looks good.

Issue #1126, opened 2024-06-10
DOI extraction 


lfoppiano:
I've noticed that there are some cases where the DOI is correctly extracted from the article header, however, they are incorrectly mangled in the output. 

Example: [origin9833693929434438741.pdf](https://github.com/user-attachments/files/15757039/origin9833693929434438741.pdf)

In this article the DOI is correctly extracted up to here: https://github.com/kermitt2/grobid/blob/694f0ed055e8c9a5d5efdc314ebef78e5e2640cf/grobid-core/src/main/java/org/grobid/core/engines/HeaderParser.java#L269 

There are several articles (e.g. Plos) where the DOI in the header is split by a breakline, in such cases the DOI is wrongly substituted. 
I'm not sure, maybe we could substitute it only when the found DOI is the same or larger than the one already extracted by the header parser. 

Another thing is that we could extract the DOI exploiting the clickable URL to correct the suffix.

lfoppiano:
I'm adding more examples (CC-BY): 
- DOI includes the final DOI: [10.1128_msphere.00700-18.pdf](https://github.com/user-attachments/files/15757266/10.1128_msphere.00700-18.pdf)
- DOI includes the page numbers [10.1073_pnas.2221103120.pdf](https://github.com/user-attachments/files/15757273/10.1073_pnas.2221103120.pdf)
- DOI is truncated, both the header parser and the regex miss the second line [10.1172_jci.insight.159058.pdf](https://github.com/user-attachments/files/15757279/10.1172_jci.insight.159058.pdf)
- Same as above: [10.1172_jci.insight.166698.pdf](https://github.com/user-attachments/files/15757298/10.1172_jci.insight.166698.pdf)

lfoppiano:
Related to the `pariod` I found a (I'm assuming) wrongly annotated training data. Adding a PR for it for version 0.8.2

lfoppiano:
Here another example: 

[huang-et-al-2019-transcriptional-sequencing-uncovers-survival-mechanisms-of-salmonella-enterica-serovar-enteritidis-in.pdf](https://github.com/user-attachments/files/17989756/huang-et-al-2019-transcriptional-sequencing-uncovers-survival-mechanisms-of-salmonella-enterica-serovar-enteritidis-in.pdf)

Even with header consolidation enabled the DOI is mangled: 

```
                    <idno type="MD5">155593D8B22A3FFE8735045743913815</idno>
[...]

PR #1127, opened 2024-06-10
Avoid replacing DOIs with shorter ones

lfoppiano:
When we search for a DOI in the page, the regex may truncate DOIs that are split by a breakline, so this PR proposes a simple fix that is to substitute the DOI only when the one found in the page is larger than the one extracted by the header parser


lfoppiano:
This fix, does not solves the main issue described in #1126 when the page are concatenated to the DOI. it sems naturally disappearing with #1142 after adding more training data and removing the search of DOI via regexes.

PR #1137, opened 2024-07-03
adding apt-get upgrade in  the builder image

vipulg13:
There are still 46 high- and 11 critical-level vulnerabilities in the "latest:crf" tag. this should apply the patches for these vulnerabilities.


lfoppiano:
Thanks!

PR #1138, opened 2024-07-05
Update docker runtime with the latest openjdk:24-slim image tag

vipulg13:
the openjdk:17-slim image is outdated with high- and critical-level vulnerabilities. Updating it with the latest image should fix the issue.

### DIFF
--- a/Dockerfile.crf
+++ b/Dockerfile.crf
@@ -63,7 +63,7 @@ RUN rm -rf grobid-source
 # -------------------
 # build runtime image
 # -------------------
-FROM openjdk:17-slim
+FROM openjdk:24-slim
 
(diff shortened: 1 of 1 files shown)


kermitt2:
Careful, we had an eclipse-temurin JDK which resulted random crashes with the docker image, I think we need to use an ubuntu official build for the docker image, because the base image for tensorflow is ubuntu-based, and we never had problems with openjdk in contrast.

lfoppiano:
@vipulg13 I'm going to close this, JDK 21 will be Long Term Support for quite some time, I would keep patching until we switch to JDK 25. https://adoptium.net/temurin/releases/?version=21&os=any&arch=any

PR #1181, opened 2024-10-11
Add includeRawCopyrights in the UI

annelhote:


lfoppiano (changes requested):
Hi @annelhote and thanks for this PR. I've noticed a problem in the XML output that is not related to this PR, that should be included. 

The licence/copyright additional XML tags added in output make the full document invalid because here is a `<p type="raw" `that is closed with a `</note>`. 

Should be fixed [here](https://github.com/kermitt2/grobid/blob/d2f0cdca8d0259c9579c04d04fa67171a69a8f38/grobid-core/src/main/java/org/grobid/core/document/TEIFormatter.java#L298) and [here](https://github.com/kermitt2/grobid/blob/d2f0cdca8d0259c9579c04d04fa67171a69a8f38/grobid-core/src/main/java/org/grobid/core/document/TEIFormatter.java#L316). 

Thank you! 😄

annelhote:
Thank you so much for your help :pray: .
Sorry about the mess in the commits, I am not (yet) used to work trough PRs but happy to learn. I'll try to remember to create dedicated branch for each PR next time.

lfoppiano (approved):
Hi @annelhote , thanks for the update. The change looks good, before merging, however, I would like to validate with @kermitt2 if the outer tag should be a `p` or a `note`.

This is the current version 
```xml
<p type="raw">Copyright © 2024 Laetitia et al. This is an openaccess article distributed under the terms of the Creative Commons Attribution 4.0 International license.</p>
```

kermitt2:
Hello, it should appear in a <p> according to https://www.tei-c.org/release/doc/tei-p5-doc/en/html/ref-availability.htmlhttps://www.tei-c.org/release/doc/tei-p5-doc/en/html/ref-availability.html 
Thanks !

lfoppiano:
Thanks! 

About the attribute `@type`? It does not seems to be expected for a `<p>` do we still want to keep it? https://www.tei-c.org/release/doc/tei-p5-doc/en/html/ref-p.html 🤔

kermitt2:
In TEI, `@type`can be added to any elements. In this case we identify the source of these copyright information similarly as the other "raw" strings.

PR #1188, opened 2024-10-23
Tensorflow 2.16

kermitt2:
This is a working PR for supporting DeLFT upgrated to tensorflow 2.16. 

This is an heavy change because this version of tensorflow is moving to Keras 3.0 - we need to use the legacy 2.0 mode, and the existing DL models are not compatible with the new TF version, and need to be retrained.


kermitt2:
See https://github.com/kermitt2/delft/pull/170 for the corresponding DeLFT upgrade

PR #1190, opened 2024-10-24
Fix URL extraction when the regex falls short

lfoppiano:
This PR fixes the URL extraction when the regular expression is shorter than the actual target (the annotated URL).


lfoppiano:
Added a fix for the case as such: 

image

Where genius editors are adding the `-` for breaking up an URL over two lines.

Issue #1192, opened 2024-10-28
Correction of idno position under analytics


lfoppiano:
Following the [question on pub2tei](https://github.com/kermitt2/Pub2TEI/issues/23) I'm adding this issue for the change discussed in there: 
 - copy the `<idno type....` under `<analytics>`
 - leave a `deprecationLocation` notice on the older elements

PR #1193, opened 2024-10-28
move TEI idno identifiers under <analytics>

lfoppiano:
See #1192 . The same treatment is applied to any identifier: PMCID, PMID, halID, etc

See example: 

```
                                </address>
                            </affiliation>
                        </author>
                        <title level="a" type="main">Transgressive phenotypes from outbreeding between the Trichoderma reesei hyper producer RutC30 and a natural isolate</title>
                        <idno type="DOI">10.1128/spectrum.00441-24</idno>
                    </analytic>
                    <monogr>
                        <imprint>
                            <date type="published" when="2024-08-20">20 August 2024</date>
                        </imprint>
                    </monogr>
                    <idno type="MD5">9E9A05DAEBD10C49EB098AF73FA55CD1</idno>
                    <idno type="DOI" status="deprecatedLocation">10.1128/spectrum.00441-24</idno>
                    <note type="submission">Received 22 February 2024 Accepted 3 July 2024</note>
                </biblStruct>
```

### DIFF
--- a/grobid-core/src/main/java/org/grobid/core/document/TEIFormatter.java
+++ b/grobid-core/src/main/java/org/grobid/core/document/TEIFormatter.java
@@ -482,6 +482,42 @@ public class TEIFormatter {
             // if it's not something in English, we will write it anyway as note without type at the end
         }
 
+        if (!StringUtils.isEmpty(biblio.getDOI())) {
+            String theDOI = TextUtilities.HTMLEncode(biblio.getDOI());
+            if (theDOI.endsWith(".xml")) {
+                theDOI = theDOI.replace(".xml", "");
+            }
+            tei.append("\t\t\t\t\t<idno type=\"DOI\">" + TextUtilities.HTMLEncode(theDOI) + "</idno>\n");
+        }
+
+        if (!StringUtils.isEmpty(biblio.getHalId())) {
+            tei.append("\t\t\t\t\t<idno type=\"halId\">" + TextUtilities.HTMLEncode(biblio.getHalId()) + "</idno>\n");
+        }
+
+        if (!StringUtils.isEmpty(biblio.getArXivId())) {
+            tei.append("\t\t\t\t\t<idno type=\"arXiv\">" + TextUtilities.HTMLEncode(biblio.getArXivId()) + "</idno>\n");
+        }
+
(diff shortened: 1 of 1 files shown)


kermitt2:
Hi Luca !

Thinking about it a second time, it might be more complicated than that and I think I understand the motivation for letting the identifiers under <biblStruct> in the case of Grobid. 

If we extract a raw DOI from a PDF, the position of the DOI in normal TEI depends on the type of document we process, either under <analytic> if this is a part of a monograph or journal, or under <monograph> if we have a report, standalone article, etc. As we can't know for sure about the document type, it make sense to put it under <biblStruct> in Grobid, to avoid errors, as a relaxed rule. 

Second, we have HAL ID which corresponds to standalone document, same for arXiv ID. We cannot put them under <analytic>, but it does not make sense to put it under <monograph> neither. <biblStruct> is a valid default choice that explicitly capture the fact that we cant know the level of the document. 

The idea behind this is: when we don't know the level associated to a extracted ID, we explicitly put it under <biblStruct>. 

@laurentromary for comments :)

kermitt2:
Note to myself: check how the DOI/ids are positioned in the case of consolidated header. We can have 2 different DOIs, one for the part and one for the "hosting" document.

lfoppiano:
@laurentromary I'm revamping this PR and I'm wondering if you have any further comment.

One option forward is to move identifier on conditions: 
- **Container detected** (journal article, book chapter, proceedings paper): article-level ids — DOI, PMID, PMCID, PII, istexId — go under `<analytic>`, because they identify the part (the article), while the journal's own ids (ISSN) already sit in `<monogr>`. 
- **No container** (preprint, report, thesis-like standalone): the same ids go under `<monogr>`, because there is no "part vs. container" split — the document is the monograph. 
- halId, arXiv, ark: these identify a standalone deposit by nature. They'd stay at the biblStruct level, unmarked (no deprecatedLocation).

This has the disadvantage of being more difficult to parse and possibly taking years for downstream process to adapt.

laurentromary:
Oui, c'est le plus cohérent. Les rapports et les thèses sont effectivement des objets monographiques. Pour les preprint, c'est moins clairs, mais c'est mieux de rester dans le même cadre (comme tu le proposes).

PR #1202, opened 2024-11-21
Alternative articles processing flavors 

lfoppiano:
This PR implements two alternatives segmentation flavors for the scientific articles: 
- `article/light`: Segment the document into header and body, and extract only title, authors, DOI, and publication date if available, body is segmented in paragraphs only.
- `article/light-ref`: Segment the document into header, body and references, and extract title, authors, DOI, and publication date if available, body is segmented in paragraphs only, references are processed as usual.

~~The article's body is then composed by two paragraphs:~~ 
~~The first paragraph contains leftover from the header, since the extraction may be sparse, this avoids to miss data~~
~~The second paragraph contains the full body~~
The article body is segmented into head and paragraphs. Tables and figures are embedded into the paragraphs too. 

The PR #1151 was tested in this PR.


lfoppiano:
I added end to end evaluation score for the new flavors: https://github.com/kermitt2/grobid/pull/1202/commits/c09aba1dea3c5a356736ce5b66597aec8b95f05f 

The scores are limited to the title, authors, first authors for the header and citations (as in the grobid normal processing) to make sure we don't introduce regressions (we shouldn't but we never know) 🙂

lfoppiano:
[here](https://grobid.readthedocs.io/en/feature-segmentation-light/Grobid-specialized-processes/) the updated documentation page with all the evaluation

PR #1227, opened 2025-01-11
Fix for CVE in CommonsIO

kermitt2:
See https://github.com/kermitt2/grobid/security/advisories/GHSA-4r7x-rm2j-vg2h


salceson:
Looks good now, thanks for your work :+1:

PR #1228, opened 2025-01-11
Updated Grobid lucene analyzers for CJK languages

kermitt2:
Use a repackaged and cleaned Lucene CJK analyzers for GROBID.


salceson:
In JAR the version of commons-io is now 2.14.0 which is OK, but in the Gradle def file it's still 2.7, which has another vulnerability. I believe that rebasing will fix the issue as in #1227 we updated it to 2.14.0.

salceson:
Now it looks OK to me, thanks!

PR #1254, opened 2025-02-25
Fix some segmentation training data.

haydn-jones:
Found some training data that was incorrectly assigning `listBibl` to body elements, might be contributing to the issues I'm experiencing.


lfoppiano:
Hi @haydn-jones, thanks for the PR. 

These training data are for the lightweight models, `article-light`, indeed the references are supposed to be considered `<body>`, however this is done in the XML parser 😅 so there is no need to fix this training data, actually.

haydn-jones:
Ah I see. Half of the files are in the standard segmentation training set though, right?

lfoppiano:
Yes, I prefer to annotate them with the full segmentation approach, expecially for what concern headnote and footnote, because we might change the lightweight model in the future. I suggest you to focus only on the data into `grobid-trainer/resources/dataset/segmentation/corpus`

haydn-jones:
@lfoppiano Sounds good, I reverted the changes for the lightweight models.

PR #1261, opened 2025-03-03
address vulnerabilities in CRF image

jgoodall:
We are trying to eliminate or reduce vulnerabilities as reported by trivy. This PR replaces the deprecated `openjdk` docker base image with `eclipse-temurin` and updates patch releases in gradle build config. The only exception (updated minor release instead of patch release) is updating `jruby-complete` to `9.4.12.0` to fix `snakeyaml` vulnerability. 

Tested locally and seemed to work. This removed all high and critical vulnerabilities from trivy:

```
grobid/grobid:0.8.2-SNAPSHOT (ubuntu 24.04)

Total: 0 (HIGH: 0, CRITICAL: 0)
```


lfoppiano:
@jgoodall I noticed that you remove TINY, usually it's used to avoid zombie processes accumulating in the docker image, but maybe is not needed anymore, do you remember what was it for a specific reason?

Issue #683, opened 2020-12-23
Only first line of figure description extracted if distance between lines deemed too large


de-code:
After having re-trained the `fulltext` and `figure` GROBID models, I noticed the extracted figure description / caption may sometimes only contain the first line despite the (DL) model correctly tagging the full description.

`482984v1` (DOI: [10.1101/482984](https://doi.org/10.1101/482984)) (from the bioRxiv 10k validation set, not to be added to training)

<details>
<summary>PDF</summary>

![image](https://user-images.githubusercontent.com/1016473/102130097-4cae7080-3e48-11eb-8af8-53d4d5f36fda.png)
</details>

<details>
<summary>target xml</summary>

```xml
<fig xmlns:mml="http://www.w3.org/1998/Math/MathML" xmlns:xlink="http://www.w3.org/1999/xlink" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" id="fig1" position="float" fig-type="figure">
<label>Figure 1.</label>
<caption><title>Regression analysis of mesophyll conductance (<italic>g</italic><sub>m</sub>) with photosynthetic-N use efficiency (PNUE) in four species seedling leaves. The determination coefficient (<italic>R</italic><sup>2</sup>) and <italic>P</italic>-value are also shown. The lines fitted separately for four species were significantly different (<italic>P</italic> < 0.05) according to the result of a one-way ANCOVA with PNUE as a dependent variable, tree species as fixed factors, and <italic>g</italic><sub>m</sub> as a covariate.</title></caption>
<graphic xlink:href="482984_fig1.tif"/>
</fig>

<fig xmlns:mml="http://www.w3.org/1998/Math/MathML" xmlns:xlink="http://www.w3.org/1999/xlink" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" id="fig2" position="float" fig-type="figure">
<label>Figure 2.</label>
[...]

PR #1288, opened 2025-05-11
Bugfix/adjust block distance when multi columns

lfoppiano:
There is a piece of code:

https://github.com/kermitt2/grobid/blob/92ea31edc391c56f6fd1eab61a16aaab5fda960c/grobid-core/src/main/java/org/grobid/core/document/Document.java#L1276

that decided whether blocks around a figure caption (something starting with fig, and tagged as `<figure>` is "close" enough to it (15px fixed distance) so that consider it or not as part of the figure. Before #1266, certain part of the document were discarded, however now are added as paragraphs. 

Now, this works in many cases, but it relies on the fact that the blocks distance is consistent. 

in this [article](https://doi.org/10.1038/s41587-025-02565-4), for example, the figure caption blocks are correctly recognized: 

Figure 1 is correctly assembled: 

image

Figure 2 is not, because the blocks dimensions are too "far" (15.16px > 15px): 

image

I did relax the rule of the 15px when the two blocks are at the same hight, but is not a solution that will cover all cases. 

any comment is welcome 😄

### DIFF
--- a/grobid-core/src/main/java/org/grobid/core/document/Document.java
+++ b/grobid-core/src/main/java/org/grobid/core/document/Document.java
@@ -1261,7 +1261,9 @@ public class Document implements Serializable {
     /**
      * This method assigns graphic objects to figures based on the proximity of the graphic object to the figure caption.
      * In addition, it removes blocks of layout tokens that are at a distance greater than a threshold from the figure caption.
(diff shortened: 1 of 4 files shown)


lfoppiano:
Here a better solution: https://github.com/kermitt2/grobid/issues/683#issuecomment-751709249

Issue #1262, opened 2025-03-04
Address vulnerabilities in docker images


jgoodall:
We are trying to eliminate or reduce vulnerabilities as reported by trivy.  I created a [PR](https://github.com/kermitt2/grobid/pull/1261) to eliminate all of the high and critical vulnerabilities reported by trivy. I tested locally and for my use case, everything worked as before.

lfoppiano:
Thanks @jgoodall I will review this in the following weeks, this PR requires a bit of testing to avoid regressions as many thing could break. May take some time to merge, sorry about that.

jgoodall:
Let me know if I can do anything to help testing. I have not seen any issues in local testing with our data.

lfoppiano:
Would be possible to add also fixes on the DL image? I'm not sure there is much that we can do given the need to use an old tensorflow version (the PR for updating is #1188). 

Regarding the tests, would be possible to process a large amount of PDF documents (e.g. in the order of thousand) and let me know if there are any problems? Basically report all the 500 that are coming out.

jgoodall:
@lfoppiano  -  For the DL image, do you mean use eclipse temurin packages instead of openjdk ones in the tensorflow image? I can try that.

Regarding tests, yes - do you have a corpus of PDF files to use for testing?

lfoppiano:
I think we need to keep the same base image, but probably installing an updated JDK. 

You could start with the dataset that are discussed [here](https://grobid.readthedocs.io/en/latest/Benchmarking/) but if you have any large set of PDF documents, that would work too.

jgoodall:
Both the [CRF](https://github.com/jgoodall/grobid/blob/master/Dockerfile.crf) and [DELFT](https://github.com/jgoodall/grobid/blob/master/Dockerfile.delft) images now use `eclipse-temurin` for java. The DELFT image still uses the old tensorflow base image for the runtime.

[...]

PR #1294, opened 2025-05-26
updated eclipse-temurin docker images to 17.0.15_6

jgoodall:
See #1262


lfoppiano:
Thanks! I think we can merge this. 

I'm also trying to get a trivy scanning directly via Github Action so that I might get early warnings. See #1295

CarAuto34:
👍

PR #1304, opened 2025-06-18
Revise Docker image for the E2E evaluation

lfoppiano:
The e2e docker image evaluation was buggy. This PR should fix it making it fully dependent on the source revision


lfoppiano:
Still the header results for authors with `strict matching` is 10% f1 score lower than previous results. I'm not sure it's biblio-glutton consolidation that has some sort of regression o somewhere else..

PR #1310, opened 2025-07-07
Update documentation

lfoppiano:
This PR aims to make the documentation less scary for python and young developers 🐉


lfoppiano:
One comment is to streamline the docker image naming, which is pretty confusing. 

We could use 
1. `grobid/grobid:{version}-{full|crf}` 
2. or we could use `lfoppiano/grobid:{version}-{full|crf}` which would be easier for me to maintain :-)

PR #1319, opened 2025-07-30
Add Conflict of interests / Author contributions

lfoppiano:
TBA


lfoppiano:
@kermitt2 it seems that the default "recipe" for training the `header-BidLSTM_ChainCRF_FEATURES` does not yield the same results as the current header model in the `master` repository. Do you remember by any chance which parameters did you use? 🙏

lfoppiano:
> @kermitt2 it seems that the default "recipe" for training the `header-BidLSTM_ChainCRF_FEATURES` does not yield the same results as the current header model in the `master` repository. Do you remember by any chance which parameters did you use? 🙏

FYI I've solved the issue and obtained similar, or slighly better, results.

PR #1341, opened 2025-10-11
Add citation.cff and update SWID 

lfoppiano:
@kermitt2 ... after the workshop on software in science, I've got motivated to update the references :-) 

Could you please quickly check whether those changes are correct?

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -144,9 +144,9 @@ JetBrains provided us with a free licence for the development:
 
 ## How to cite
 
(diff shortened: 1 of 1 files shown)


kermitt2:
Looks good! But  I would change:

```
message: "If you use this software, please cite it as below."
```

by 

```
message: "If you want to reference this software, please cite it as below."
```

because with Apache2 License, there is no obligation of disclosure or citation when using the software.

lfoppiano:
Thanks @kermitt2

PR #1348, opened 2025-11-08
hotfix: fixed docker build crash due to dynamic Python env version fetching

thesanogoeffect:
Hi,

Running `docker build -t grobid/grobid:0.8.3-SNAPSHOT --build-arg GROBID_VERSION=0.8.3-SNAPSHOT --file Dockerfile.delft .` (as per https://grobid.readthedocs.io/en/latest/Grobid-docker/) resulted in the following error:

124 | 125 | # Get the Python version dynamically and set LD_LIBRARY_PATH accordingly 126 | >>> ENV PYTHON_VERSION=$(python3 --version | cut -d' ' -f2 | cut -d'.' -f1,2) 127 | ENV LD_LIBRARY_PATH=/usr/local/lib/python${PYTHON_VERSION}/dist-packages/jep:grobid-home/lib/lin-64:grobid-home/lib/lin-64/jep:${LD_LIBRARY_PATH} 128 | # remove libjep.so because we are providing our own version in the virtual env above -------------------- ERROR: failed to build: failed to solve: Syntax error - can't find = in "--version". Must be of the form: name=value

This was due to the incorrect dynamic Python version ENV fetching in Dockerfile.delft (the same issue was in Dockerfile.evaluation). This doesn't work in Docker because ENV cannot execute shell commands.

According to git history, this change was introduced in e20e653dd11f1465ddcd537cf4855aade5297405 and included in yesterday's homfunc merge. 

After considering how to fix this approach, I reverted back to effectively hardcoding the Python version, which seems cleaner here. I've also ensured that all ENV variables follow the key=value syntax to avoid warnings and linter messages.

Best,
Jakub

### DIFF
--- a/Dockerfile.crf
+++ b/Dockerfile.crf
@@ -52,7 +52,7 @@ RUN rm -rf grobid-home/lib/mac-64
 # cleaning Delft models
 RUN rm -rf grobid-home/models/*-BidLSTM_CRF*
 
-ENV GROBID_SERVICE_OPTS "-Djava.library.path=grobid-home/lib/lin-64:grobid-home/lib/lin-64/jep"
+ENV GROBID_SERVICE_OPTS="-Djava.library.path=grobid-home/lib/lin-64:grobid-home/lib/lin-64/jep"
 
 RUN ./gradlew clean assemble --no-daemon  --info --stacktrace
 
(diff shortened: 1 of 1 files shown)


lfoppiano:
Thanks @thesanogoeffect!

PR #1439, opened 2026-05-06
Development / Model Debug API

### LINKED ISSUE #1435: Development/Debug API mode
Add an API for development, that allow the requester to obtain the tagging output from the different models. 

Add a parameter to the current server paths, called `debug_mode=true`: 
- where each model raw annotations will be returned instead of the XML TEI
- however a parameter `models=model1, model2` would return only specific models output

lfoppiano:
This PR introduces a development or model debug API. 

Through this API is possible to get the model output from a single or multiple models via API. This is a great improvement because it helps assessing whether the bug is from the code or from the data, without debugging Grobid.


de-code:
Seems to be working as expected

PR #1524, opened 2026-08-06
Handle middle dot (U+00B7) as bullet only outside words

toniher:
Remove middle dot from the generic bullet_chars pattern and add a dedicated pattern that treats it as a bullet only when not adjacent to word characters. This preserves intra-word uses like Catalan "punt volat" (intel·ligència), chemical formulas (CuSO4·5H2O), and mathematical notation, while still normalizing middle dots used as list markers to the standard bullet character (U+2022).

Disclaimer: Claude Opus 5.0 support

### DIFF
--- a/grobid-core/src/main/java/org/grobid/core/utilities/UnicodeUtil.java
+++ b/grobid-core/src/main/java/org/grobid/core/utilities/UnicodeUtil.java
@@ -138,9 +138,20 @@ public class UnicodeUtil {
             + "\\u25CF"  // black circle
             + "\\u26AB"  // medium black circle
             + "\\u2B24"  // black large circle
-            + "\\u00B7"  // middle dot
             + "]";
     private static final Pattern BULLET_CHARS_PATTERN = Pattern.compile(bullet_chars);
+
+    // U+00B7 MIDDLE DOT is deliberately NOT part of bullet_chars above, because it is
+    // ambiguous: it is a common list bullet, but it is also a letter-level character
+    // *inside* words. The main case is the Catalan "punt volat" separating a geminate l,
+    // as in "intel·ligència" or "cel·lular", but it also occurs in chemical formulas of
+    // hydrates ("CuSO4·5H2O") and as a multiplication/dot-product sign. Rewriting it to
+    // a bullet there corrupts the token and, because U+2022 is one of
+    // TextUtilities.delimiters while U+00B7 is not, additionally splits the word into
+    // three tokens ("intel", "•", "ligència"), degrading every downstream model.
+    // So only treat a middle dot as a bullet when it is not between two word characters.
+    private static final Pattern MIDDLE_DOT_BULLET_PATTERN = Pattern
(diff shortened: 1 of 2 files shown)


lfoppiano:
Hi @toniher ! Thanks for the PR. Could you please share also a couple of examples of PDF with this issue, so that I can test them? 
Thank you

toniher:
Hi @lfoppiano ! Thanks for considering it. Any Catalan text would be a good case. I first noticed it here: https://zenodo.org/records/13836801/files/GuiaPublicacio-Zenodo-Article-Catalan-v1.pdf?download=1 , but any other PDF from that community would work, since it is a common situation: https://zenodo.org/communities/recercatala/records . Another one, for instance: https://zenodo.org/records/15261991/files/GCATSSeldem-SCIRE.Article-Catalan-v1.pdf?download=1

I have not managed to find cases for other alike usages, such as hydrates — yet. It appears that authors try to (understandably) avoid mid-dots: https://pubmed.ncbi.nlm.nih.gov/?term=CuSO4%C2%B75H2O

lfoppiano:
I ran a test on this paper: https://zenodo.org/records/13836801/files/GuiaPublicacio-Zenodo-Article-Catalan-v1.pdf?download=1

- this PR: `llenguatge, de la intel•ligència artificial generativa` 
- [last 0.9.1](https://huggingface.co/spaces/grobidOrg/grobid-full): `llenguatge, de la intel•ligència artificial generativa`

So this is correct, they should be preserved. Right? 

Then I have `• acostar els resultats de la recerca als ciutadans que són els que, en últim terme` which is part of the bullet points, which in both grobid output (this PR and the last version) is the same character. 🤔  I'm not understanding which differences should I expect?

toniher:
Hi @lfoppiano 
Just tested this branch against 0.9.1 and I get this.

---

Key finding: this confirms the bug your branch fixes. GROBID 0.9.1 corrupts the Catalan middle dot (U+00B7) into a bullet character (U+2022):

- 0.9.1 (buggy): intel•ligència
- This branch (bug/middot, fixed): intel·ligència

---

Let me share you the output files as well. 

It is strange that you get the same in both cases. Could you double-check whether you tested this PR branch?

Thanks!

[test.txt](https://github.com/user-attachments/files/31694935/test.txt) (this branch)
[test-0.9.1.txt](https://github.com/user-attachments/files/31694936/test-0.9.1.txt) (0.9.1 release)

lfoppiano:
@toniher thanks! I've checked again and the PR looks good. 

I've added an additional safe on the CJK languages and some more tests. For merging you should:

0. make sure the Github Actions are running on your fork 
1. sync your fork's master, 
2. rebase this branch onto master and 
3. force-push here

Then we should be able to run the CI and then merge 🤞

lfoppiano (approved):
LGTM! 👍

lfoppiano:
@toniher probably nothing to do on your side, the CI are running.

lfoppiano:
Moltas gracias @toniher  😃

PR #1557, opened 2026-09-10
Add citation training data for bracketed medium labels before a URL

### LINKED ISSUE #1556: Closing ] is dropped from a title when the reference ends with a URL
### Grobid version

GROBID 0.9.1

### Operating System and architecture (arm64, amd64, x86, etc.)

Mac OS, docker image

### What is your Java version

_No response_

### Log and information

_No response_

### Further information

This is my first engagement with the GROBID community. Thank you so much for all the work done on this software.

### Summary

When a reference ends with a URL, a closing square bracket immediately before that URL is lost from
the extracted title. A closing round bracket in the same position is kept, and the same square
bracket is kept when the reference ends with anything other than a URL.

### To reproduce

```bash
curl -s -X POST http://localhost:8070/api/processCitation \
  --data-urlencode 'citations=Smith, J. (2020). Alpha [Computer software]. https://doi.org/10.5281/zenodo.1234567'
```

### Result

```xml
<title level="m">Alpha [Computer software</title>
```

### Expected

```xml
<title level="m">Alpha [Computer software]</title>
```

### Minimal pairs

Each row differs from the first in one respect only.

| Input (after `Smith, J. (2020).`)                     | Extracted title                | `]` / `)` |
| ----------------------------------------------------- | ------------------------------ | --------- |
| `Alpha [Computer software]. https://doi.org/10.5281/zenodo.1234567` | `Alpha [Computer software` | **lost**  |
| `Alpha (Computer software). https://doi.org/10.5281/zenodo.1234567` | `Alpha (Computer software)` | kept      |
| `Alpha [Computer software]. https://example.org/alpha`              | `Alpha [Computer software` | **lost**  |
| `Alpha [Computer software]. Some Publisher.`                        | `Alpha [Computer software]` | kept      |
| `Alpha [Computer software]. doi:10.5281/zenodo.1234567`             | `Alpha [Computer software]` | kept      |

So it is specific to `]`, and specific to a URL following it — not to DOIs (the bare `doi:` form is
unaffected) and not to closing punctuation in general.

### Why it matters

`[Computer software]`, `[Dataset]`, `[Preprint]` and similar medium labels are how APA and several
other styles mark non-article works, and such references very often end with a URL. Downstream consumers get a title with an unbalanced
bracket, which does not match the same work as recorded anywhere else.

---

## Secondary, and less clear-cut: software citations lose their title entirely

Same endpoint and version. This one has no single-variable reproduction, so it may be a training-data gap
rather than a defect.

APA software citations are parsed inconsistently, and the outcome changes with the number of authors
while nothing else about the reference changes. Holding the rest of the string fixed as
`SolrWayback (Version 5.4.1) [Computer software]. https://doi.org/10.5281/zenodo.18314399` and
varying only how many authors precede it:

| Authors | `<title>`                       | `<note>`                                       |
| ------- | ------------------------------- | ---------------------------------------------- |
| 1       | `SolrWayback (Version 5.4.1) [Computer software` | —                             |
| 2       | `SolrWayback`                   | `Version 5.4.1) [Computer software`            |
| 3       | `SolrWayback (Version 5.4.1)`   | `Computer software`                            |
| 4       | `SolrWayback (Version 5.4.1)`   | `Computer software`                            |
| 5–8     | *(empty)*                       | `SolrWayback (Version 5.4.1) [Computer software` |

At five authors and above the title element is empty and the whole title is in an untyped `<note>`.
The same reference with `Royal Danish Library.` in place of the URL is parsed correctly at every
author count, so the trailing URL is again part of it.

The title string matters too, independently of author count. With two authors,
`InternetArchiveExtractor (Version 0.0.8) [Python]. https://doi.org/...` yields an empty title and
`InternetArchiveExtractor (Version 0.0.8) [Python` as a note, and it still does with one author.

Reproduction script:

```bash
for n in 1 2 3 4 5 6 7 8; do
  authors=$(python3 -c "
names=['Egense, T.','Eskildsen, T.','Lauridsen, J.',\"O'Brien, B.\",'Jackson, A.','Bellony, L.','Thogersen, J.','Johnston, V. H.'][:$n]
print(', '.join(names[:-1]) + ', & ' + names[-1] if len(names) > 1 else names[0])")
  echo "--- $n author(s) ---"
  curl -s -X POST http://localhost:8070/api/processCitation \
    --data-urlencode "citations=$authors (2026). SolrWayback (Version 5.4.1) [Computer software]. https://doi.org/10.5281/zenodo.18314399" \
    | tr -d '\n' | grep -o '<title[^>]*>[^<]*</title>\|<note[^>]*>[^<]*</note>'
done
```

abeeranajam31:
## What

Adds 19 hand-annotated `<bibl>` training examples to
`grobid-trainer/resources/dataset/citation/corpus/` covering APA-style
software/dataset citations where a bracketed medium label (e.g.
`[Computer software]`, `[Dataset]`, `[Preprint]`) is immediately
followed by a URL, with no `[Online]. Available:` phrase in between.

## Why

#1556 reports that GROBID drops the closing `]` from a citation title
when the reference ends with a URL, e.g.:

```
Smith, J. (2020). Alpha [Computer software]. https://doi.org/10.5281/zenodo.1234567
```

extracts `<title level="m">Alpha [Computer software</title>` instead of
`...software]</title>`. The issue also documents a related failure
where the title disappears entirely as author count increases, for the
same "bracket immediately before a URL" pattern.

I traced this through `CitationParser.resultExtractionLayoutTokens`
(grobid-core): the title text is built directly from the citation
model's per-token cluster labels (`clusterContent` from
`clusteror.cluster()`), with no string post-processing in between. So
this isn't a Java-side bug to patch — the label boundary between
`<title>` and `<ptr type="web">` is learned entirely from the training
corpus.

I checked the existing corpus for this pattern: there are ~29 places
where `]` sits right before a `<ptr type="web">`, but in every one I
found, the bracket closes an `[Online]` aside (untagged transition
text before "Available:"), never a bracketed label that is part of the
title itself with no such phrase in between. That gap is a plausible
explanation for the model treating the bracket as belonging to the
URL/transition zone rather than the title when there's no `[Online]`
cue.

## What this doesn't do

I don't have Java/Gradle available in the environment I used to put
this together, so I could not run this through
`CitationTrainer`/`generateTrainingData` or retrain the model to
confirm the fix end-to-end. The XML is well-formed and follows
`doc/training/Bibliographical-references.md` (checked against many
existing corpus examples for tag conventions, e.g. `doi.org` URLs are
tagged `<ptr type="web">` throughout the corpus, not `<idno type="DOI">`,
which I matched here too). Happy to adjust the annotations if I've
misread a convention, or to add more variants if useful — this doesn't
close #1556 on its own since the shipped model needs a retrain to pick
it up.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/grobid-trainer/resources/dataset/citation/corpus/software_dataset_citations_batch.xml
+++ b/grobid-trainer/resources/dataset/citation/corpus/software_dataset_citations_batch.xml
@@ -1,24 +1,13 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <tei xmlns="http://www.tei-c.org/ns/1.0" xmlns:xlink="http://www.w3.org/1999/xlink" xmlns:mml="http://www.w3.org/1998/Math/MathML">
 <listBibl>
-<bibl><author>Smith, J.</author> (<date>2020</date>). <title level="m">Alpha [Computer software]</title>. <ptr type="web">https://doi.org/10.5281/zenodo.1234567</ptr></bibl>
-<bibl><author>Smith, J.</author> (<date>2020</date>). <title level="m">Alpha [Computer software]</title>. <publisher>Zenodo</publisher>.</bibl>
-<bibl><author>Smith, J.</author> (<date>2020</date>). <title level="m">Alpha [Computer software]</title>. <idno type="DOI">doi:10.5281/zenodo.1234567</idno></bibl>
 <bibl><author>Egense, T.</author> (<date>2026</date>). <title level="m">SolrWayback (Version 5.4.1) [Computer software]</title>. <ptr type="web">https://doi.org/10.5281/zenodo.18314399</ptr></bibl>
 <bibl><author>Egense, T. &amp; Eskildsen, T.</author> (<date>2026</date>). <title level="m">SolrWayback (Version 5.4.1) [Computer software]</title>. <ptr type="web">https://doi.org/10.5281/zenodo.18314399</ptr></bibl>
 <bibl><author>Egense, T., Eskildsen, T. &amp; Lauridsen, J.</author> (<date>2026</date>). <title level="m">SolrWayback (Version 5.4.1) [Computer software]</title>. <ptr type="web">https://doi.org/10.5281/zenodo.18314399</ptr></bibl>
 <bibl><author>Egense, T., Eskildsen, T., Lauridsen, J. &amp; O'Brien, B.</author> (<date>2026</date>). <title level="m">SolrWayback (Version 5.4.1) [Computer software]</title>. <ptr type="web">https://doi.org/10.5281/zenodo.18314399</ptr></bibl>
 <bibl><author>Egense, T., Eskildsen, T., Lauridsen, J., O'Brien, B. &amp; Jackson, A.</author> (<date>2026</date>). <title level="m">SolrWayback (Version 5.4.1) [Computer software]</title>. <ptr type="web">https://doi.org/10.5281/zenodo.18314399</ptr></bibl>
 <bibl><author>Egense, T., Eskildsen, T., Lauridsen, J., O'Brien, B., Jackson, A. &amp; Bellony, L.</author> (<date>2026</date>). <title level="m">SolrWayback (Version 5.4.1) [Computer software]</title>. <ptr type="web">https://doi.org/10.5281/zenodo.18314399</ptr></bibl>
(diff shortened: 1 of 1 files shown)


lfoppiano (changes requested):
@abeeranajam31 thank you! 
In order to keep track of our data, could you please provide the source where those references were collected? The direct URL or the DOI would be sufficient.

abeeranajam31:
> Perhaps for this send me an email at luca AT sciencialab _ com.

thank you for the guidance, have sent you an email

